PR #304, opened 2011-06-16
Last-child pseudo selector test

laustdeleuran:
I don't know how this is merged into Modernizr, but I've added the last-child pseudo selector test the feature-detection folder. The script has been tested in FireFox 3.0+, Chrome 7, 10, 11,12, IE6-9, Safari 4.0+ and Opera 11.

It would probably be a good idea to test the script in FireFox 1.0-2.0 and Safari 1.3-2.0. I haven't been able to perform test in those browsers.

### DIFF
--- a/feature-detects/css-lastchild.js
+++ b/feature-detects/css-lastchild.js
@@ -1,34 +1,3 @@
-// :last-child pseudo selector test.
-// By @laustdeleuran and @emilchristensen
-// Please see http://reference.sitepoint.com/css/pseudoclass-lastchild for weird :last-child behaviour in FireFox below 2.0 and Safari below 2.0.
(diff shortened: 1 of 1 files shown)


es6Test:
If I just include modernizr will :last-child just work on a ie8 stylesheet? I done that and it didn't work.

phistuck:
Modernizr does not add features to the browser, it only detects whether the features are supported.

If you selected HTML5Shiv, though, HTML5Shiv (not Modernizr) makes the browser recognize HTML5 elements as unknown elements (as opposed to empty nodes that cannot be laid out using CSS).

es6Test:
I see, I usually just do that by checking if the api is not undefined with vanillajs.

phistuck:
@es6Test - for JavaScript features, this is pretty easy, yes. For prefixed CSS properties, for example, it gets trickier.

PR #659, opened 2012-08-18
fixes #605: Clipboard API feature-detection

### LINKED ISSUE #605: copy to clipboard test
?

fivetanley:
Specification: http://www.w3.org/TR/clipboard-apis/
DataTransfer Interface: http://www.w3.org/TR/html5/dnd.html#the-datatransfer-interface
Event Interfaces: http://www.w3.org/TR/DOM-Level-3-Events/#event-interfaces

**tl;dr: It's pretty much impossible, I think, to fakeout a real `paste` event from the user's clipboard, because the spec does not allow it.  We can't check for a `clipboardData` on the `paste` event because of this. So, we need to check for the `ClipboardEvent` constructor (which is spec'd), then we can check for the existence of `clipboardData` on a new `Event` (non-standard/not in spec).**

The best way to test this feature would to do it manually by hand:

1. Attach some event listeners for a 'paste' event to a DOM element
2. Paste something by hand into that DOM element.
3. Capture the event, and see if its `clipboardData` propertyexists AND implements the `DataTransfer` Interface:

Old Interface:

* `setData` (function)
* `getData` (function)
* `clearData` (function)

And/or the new one, which looks like it is meant for drag/drop only?

* `setDragImage` (function)
* `addElement` (function)

Firefox supports `paste`, `copy`, and `cut` events, but there is no `clipboardData` property on the `Event` passed to event listeners.  Simply checking for `element.onpaste` won't work.

The best check would be a `for in` check for those properties, because they may be on the prototype chain (most likely native code). The spec for DataTransfer seems to be designed for Drag/Drop events, so the `setDragImage` function probably shouldn't *have* to implemented. `addElement` is questionable.

Chrome 21 and Safari 5.1 support the old interface, so we could test that.

Why manually test?  The specification allows for synthetic `ClipboardEvent`s, but, **does not allow for those synthetic events to read/write from the clipboard**.  They may only pass `data` and `dataType` properties on the passed `Event` object to the event listeners.  So, even if we manually dispatched events using `document.createEvent` and other methods, we couldn't mock the behavior to check for `clipboardData` on the passed `Event` object because:

* No browsers currently implement the `ClipboardEvent` constructor anyway.
* Using the `ClipboardEvent` constructor wouldn't pass the `clipboardData` property on the event, or would probably throw an exception depending on implementation.  The spec doesn't want these synthesized events to be able to read or modify clipboard data (for good reason).
* Creating a `paste` event on an object doesn't create a `clipboardData` property, and creating that property on an event would make the detection work in browsers it shouldn't.

Link to Clipboard Event Interfaces:
http://www.w3.org/TR/clipboard-apis/#clipboard-event-interfaces

Link to test: http://jsbin.com/igitud/14/edit
 
Using the test above, these browsers have been tested and found to implement the Clipboard API:

* Chrome 21
* Safari 5.1

These browsers have been tested and found to NOT implement Clipboard API:

* Firefox 14
* Firefox Aurora ( 15 ) ( [It may land soon...](https://bugzilla.mozilla.org/show_bug.cgi?id=407983) )

The test I comitted basically checks for the existence of a `ClipboardEvent` constructor, as per the spec.  Next, it checks to see if creating a new `Event` creates a `clipboardData` property on the created object; This may be a WebKit-specific detail, because the spec does not require a `clipboardData` property for the `Event` interface.

-

fixes #605


Seebiscuit:
A couple of issues: 

1. Firefox, which implements `clipboardData.setData()`/`clipboardData.getData()` and also implements the `ClipboardEvent` constructor, does not actually modify the user's clipboard on `setData()`. AFAIK, IE and Chrome do. How can we feature-detect if we can modify the user's clipboard?

2. @fivetanley PR checks for the 'paste' event. Should it also, at least, check for 'copy'?

hallvors:
Hi, I'm editing the Clipboard Events spec and apologise for the poor feature detection story - we've been trying to figure this out for a while..

Some random responses to above comments:

@fivetanley says

> Using the ClipboardEvent constructor wouldn't pass the clipboardData property on the event,
> or would probably throw an exception depending on implementation. 

Using the ClipboardEvent constructor should certainly create an event with a clipboardData property. (If you see an implementation where this doesn't happen, I hope you report a bug..)

For example, the sample script in the spec - currently example 4:
https://w3c.github.io/clipboard-apis/#clipboard-event-interfaces
would trigger a paste event listener, and calling event.clipboardData.getData('text/plain') from within that event listener should return "My string". 

(However, if you called the constructor without any data, a synthetic event would *not* actually show you data from the real clipboard. Also, even if you dispatched the paste event from that example to an editable element, the "My string" text would not actually get inserted, since a synthetic event doesn't have a default action.)

@Seebiscuit :

> How can we feature-detect if we can modify the user's clipboard?

Sorry, you can't. And you won't be able to (unless the user grants you privileges to do document.execCommand('paste')). But I'm still trying to find a way to improve feature detection - suggestions welcome.

patrickkettner:
@hallvors thank you _so_ much for reaching out. This has been in the back of my mind for years :]

given the following

1. we must not change the user's clipboard
2. we must use a synthetic event
3. synthetic event be able to actually work (clipboard poisoning, etc)

is there any chance that a synthetic clipboard copy/paste could throw a unique err that we could `try`/`catch` around and sniff for?

patrickkettner:
(sorry for the close - cat walked on the keyboard :[ )

ryanseddon:
@patrickkettner that's a great idea, if we try to poison the clipboard with a synthetic copy event it should probably throw an `UNSTRUSTED ` error or something along those lines.

JamesMGreene:
I'm OK with that idea personally, and I'm guessing it would be easier for both browser vendors and consumers to implement than some of the other previously discussed approaches for Clipboard API feature detection (e.g. asking browser vendors to revive the mostly abandoned `document.queryCommandSupported` and `document.queryCommandEnabled` methods to return accurate results again).

@patrickkettner @ryanseddon I've cited both of your last comments in an ongoing related discussion on the public-webapps@w3.org group mailing list.

patrickkettner:
thanks @JamesMGreene!

[link for the lazy](https://lists.w3.org/Archives/Public/public-webapps/2015AprJun/0175.html)

hallvors:
@ryanseddon :
>  if we try to poison the clipboard with a synthetic copy event it should probably throw an UNTRUSTED error or something along those lines.

I think these events have use cases - for example if you have a "cloud clipboard" implementation where snippets can be stored in your web service and re-used you could fire synthetic events to make sure cut/copy/paste actions related to the cloud clipboard went through the same processing as normal actions. So I don't see synthetic events as attempts to "poison" the clipboard - I see them as a way one part of your script/site/app lets another part of your script/site/app know that there is an intent to cut, copy or paste some data.

What do you think about this approach, though?
https://lists.w3.org/Archives/Public/public-webapps/2015AprJun/0173.html

patrickkettner:
@hallvors I am not 100% sure that we are talking about the same thing.

Are you suggesting that a developer will be able to synthetically create a `cut` event and have it actually update the clipboard with zero interaction from their end user?

hallvors:
No. The syntetic event has no default action, so it will not cause the system clipboard to be updated. But if you have added cut event listeners, they would run. The script triggering the cut event can then inspect the DataTransfer object and pick up any modifications the cut listeners did. 

For example, in a UI where you move "objects" around, the cut listener might for example identify the "object" that has focus, remove it from the document and keep a reference, and call preventDefault() or return false. The event dispatching script can check the return value to know if the "cut" intention has been handled - if it hasn't, use its own logic for "cut".

fivetanley:
As evidenced by my lack of activity, I think this pull request should be closed. I don't have the time or capacity to work on it. Best of luck to those who do!

patrickkettner:
@fivetanley if you don't mind, I would like to have it open.

Now that the spec authors are included in the conversation, I think may be able to not be stuck :]

fivetanley:
Nice, cool. Glad to help facilitate discussion!

patrickkettner:
thank you so much!

On Tue, Apr 21, 2015 at 1:58 PM, Stanley Stuart <notifications@github.com>
wrote:

> Nice, cool. Glad to help facilitate discussion!
>
> —
> <https://github.com/Modernizr/Modernizr/pull/659#issuecomment-94889807>.
>



-- 
patrick

patrickkettner:
@hallvors having a listener pickup on one of the events would work fine (though that does mean it becomes an async detection, is less than ideal since it has to include a setTimeout backup to ensure listeners area cleaned up and what not...). 

Out of curiosity, why would someone want listeners to fire for synthetic events?

JamesMGreene:
I haven't tested this out to see if it dispatches all the correct events in order to be spec-compliant but Chromium just landed programmatic copy-and-cut [with restrictions that require that there is an active, expanded Selection/Range in the Document]:

http://updates.html5rocks.com/2015/04/cut-and-copy-commands

For better or worse, the time for spec changes is fleeting....

hallvors:
@patrickkettner: document.execCommand() fires clipboard events synchronously (per the spec) so no timeout is required.

> why would someone want listeners to fire for synthetic events?

Well, why would you dispatch a synthetic event if you didn't want event listeners to fire? :)

hallvors:
@JamesMGreene 
> For better or worse, the time for spec changes is fleeting....

Agreed, although the solution I prefer at the moment is sort of orthogonal to the rest of the spec since it requires most work in queryCommandEnabled() implementation. So as far as I can tell they won't have to do major rewrites of newly written code if we add feature detection that way.

patrickkettner:
@hallvors 
> document.execCommand() fires clipboard events synchronously (per the spec) so no timeout is required.

so something like the following is what you are suggesting?

```javascript
Modernizr.addTest('cilpboard', function() {
  var supported = false;
  var listener = function(e) {
    supported = true;
    e.preventDefault();
  }

  var copyEvent = new ClipboardEvent(copy ', {
    dataType: 'text/plain',
    data: 'modernizr rulez'
  });

document.addEventListened('beforecopy', listener);
document.dispatchEvent(copyEvent);
document.removeEventListened('beforecopy', listener);

return supported;
})
```

(obviously would need to have a few more checks, like checking the `ClipboardEvent` constructor exists, etc...)

Currently that is _not_ working in any browser I can find.

> Well, why would you dispatch a synthetic event if you didn't want event listeners to fire? :)

To sniff for actual support :p

In all seriousness, if I had a listener for that event, I would only expect it to fire for situations where the actual `copy` event can fire. It seems rather unintuitive currently


My concern with `queryCommandEnabled` is that it is another abstraction required. It feels really similar to `document.implementation.hasFeature`, which never really worked for almost all cases, and is possibly being dropped from chrome soon.

hallvors:
Something similar to that, but maybe more like 

```js
Modernizr.addTest('clipboard', function() {
  var supported = false;
  var listener = function(e) {
    supported = e.clipboardData && e.clipboardData.getText('text/plain') == 'modernizr rulez';
  }

  var pasteEvent = new ClipboardEvent('paste', {
    dataType: 'text/plain',
    data: 'modernizr rulez'
  });

  document.addEventListener('paste', listener);
  document.dispatchEvent(pasteEvent);
  document.removeEventListener('paste', listener);
  return supported;
})
```

Or something like this:
https://gist.github.com/hallvors/59a90f2e3816cb57f044

I haven't tested the former in any browser, I've tested the gist code a bit.

queryCommandEnabled() is already used to check if editing commands can be used - it's much "closer" to the functionality you want to investigate here than hasFeature() ever was. If you want to detect "in the instant" - for example from a mousedown event - if copy/cut are enabled in the context of that user action, it would be useful - if it isn't broken by some UAs returning false positives already. I suspect it is, but haven't tested..

hallvors:
http://jsfiddle.net/213ov2ah/ tests queryCommandEnabled() for copy and cut - non-user-initiated and user-initiated.

patrickkettner:
am I supposed to be getting false for both on chrome canary?

hallvors:
Does the build you're testing support user-triggered document.execCommand('copy') ? If yes, I'd consider it a clear bug if it says false on both counts. However, it's a fixable bug and not a false positive - which is a greater concern. False negatives here just means a few Flash click-to-copy hacks will live a little longer, while a false positive shipping in a major browser would make the feature detection approach unreliable.

patrickkettner:
> Does the build you're testing support user-triggered document.execCommand('copy') 

yesir

> I'd consider it a clear bug if it says false on both counts

That is one of the main reasons I am pushing for the throw-an-error-when-invoked-via-JS thing. There have been a number of APIs that are really just black boxes that we hope have been wired correctly to indicate if something is supported (`canPlayType`, `supportsContext `, `probablySupportsContext`, etc, etc) that have a history of at least occasionally just being completely wrong. While it makes perf sense for things that require spinning up the GPU or deciding which video to load, I really don't see the case for a similar API for execCommand (`queryCommandEnabled`)

> a false positive shipping in a major browser would make the feature detection approach unreliable

but it _is_ unreliable. In its current form. It exists - can't detect it. Obviously it would be worse if you thought it did work, but this situation is just as bad if someone is in Chrome 43+ and has flash blocked.

hallvors:
@paulirish what's the chance of getting this bug fixed before Chrome ships this stuff in a production build?

patrickkettner:
since paul has IO today, I'll jump in.

0% - it is in Chrome 43, which shipped as stable last week

hallvors:
OK. I've nagged Daniel Cheng on the Chrome team - hope they can fix it soon.

The main use case for queryCommandEnabled() is to set button state correctly when you've created your own editor UI.

patrickkettner:
I get that, it is just already inconsistent.

I am worried that what is a more or less one off API will end up like `hasfeature` or any of the other similar apis in the past. I get that there is a valid use case for it, but I would put forth that the error throwing thing makes code errors much more apparent much more quickly than just returning false.

hzoo:
Was this meant to be closed?

patrickkettner:
nope - thanks!

pixelbandito:
What's the status of this PR now? Let me know if you need encouragement. :) :fireworks: :smile:

PR #911, opened 2013-04-22
Font-Smoothing test

wellcaffeinated:
Hi there,

I've adapted [Zoltan Du Lac's font-smoothing detection](http://www.useragentman.com/blog/2009/11/29/how-to-detect-font-smoothing-using-javascript/) into a V3 test.

Thought it might be useful to submit it here.

This is my first submission to Modernizr, so any feedback/code reviews would be much appreciated.

### DIFF
--- a/feature-detects/fontsmoothing.js
+++ b/feature-detects/fontsmoothing.js
@@ -1,0 +1,103 @@
+/*!
+{
+  "name": "Font-Smoothing",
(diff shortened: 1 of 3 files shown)


patrickkettner:
Given the general lack of demand combined with the age of the issue, I am going to go ahead and offically not accept this.

Thank you so much for opening it, @wellcaffeinated !

PR #983, opened 2013-06-30
Add better async test support. Fixes #969.

### LINKED ISSUE #969: Async tests should add property to `Modernizr` object when declared
Copy-paste from my comment on #622.

It might be nice if we attach the property to the Modernizr object when async tests are declared (but leave it `undefined`) – i.e.:

```html
<script src="modernizr.js"></script>
<script>
'datauri' in Modernizr  // true (with the current code this would be false)
Modernizr.datauri       // undefined
</script>
```

in case it's ever helpful to see which tests have been defined, without them having to have finished running.

Although that might mean changing the APIs (which may be no bad thing – I've mentioned before I think it's a little bit confusing):

```javascript
Modernizr.addAsyncTest('datauri', function () {
   // ...
      resolveAsyncTest(true);
   // ...
});
```

Alternatively could have `resolveAsyncTest()` and `rejectAsyncTest()` to match Promise language (both take no arguments).

We'd keep the `addTest()` API the same as it is now; `resolveAsyncTest()` could just be a wrapper around it.

@SlexAxton – thoughts?

SlexAxton:
This probably needs a decent read-through @stucox -- but I think it's what you want. I haven't quite been bothered to rewrite the tests that this fails. They are generally expecting false values for async tests, and now getting explicit `undefined`s. But as far as I can tell, that's desired functionality now. Went through and converted all async tests to the new format.

### DIFF
--- a/feature-detects/canvas/todataurl.js
+++ b/feature-detects/canvas/todataurl.js
@@ -3,7 +3,6 @@
   "name": "canvas.toDataURL type support",
   "property": ["todataurljpeg", "todataurlpng", "todataurlwebp"],
   "tags": ["canvas"],
-  "async" : false,
   "notes": [{
     "name": "HTML5 Spec",
     "href": "http://www.w3.org/TR/html5/the-canvas-element.html#dom-canvas-todataurl"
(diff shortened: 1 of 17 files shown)


patrickkettner:
@SlexAxton I _think_ this would fix a bug I am trying to track down - any chance you could give this a quick rebase?

rejas:
Hi @SlexAxton do you have the time to rebase this against the latest master so we can have a better look at ti again?

PR #1257, opened 2014-03-09
update Modernizr._prefixes to only return the supported prefixes

### LINKED ISSUE #1223: Don’t test against every prefix every time for `testAllProps()` / `prefixed()`
As highlighted by @hexalys on #878: it’s wasteful to iterate through every prefix every time – the prefix an engine uses is always the same. IE gets a particularly raw deal, because we test `-ms-` last every time.

I also like the idea of exposing the detected prefix as `Modernizr._prefix`.

How should we determine the prefix? [This method](http://davidwalsh.name/vendor-prefix) by @darkwing looks as reliable as any other (taking @mathiasbynens’ [comment re. IE < 8 support](http://davidwalsh.name/vendor-prefix#comment-49324) into account)

patrickkettner:
fixes #1223

### DIFF
--- a/distmetadata.json
+++ b/distmetadata.json
@@ -1,0 +1,7423 @@
+[
+  {
+    "name": "Ambient Light Events",
+    "property": "ambientlight",
+    "notes": [
+      {
+        "name": "W3C Ambient Light Events",
+        "href": "http://www.w3.org/TR/ambient-light/"
+      }
+    ],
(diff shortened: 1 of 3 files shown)


patrickkettner:
gonna close this, as browsers now support a mish mash of prefixes and you can't guarantee which any browser will implement.

PR #1263, opened 2014-03-12
Content Security Policy restrictions workaround

### LINKED ISSUE #1262: Content Security Policy restrictions error
Demo

When CSP is on Modernizr can accidently crash entire app (when Modernizr comes as part of app file).

termi:
fixes #1262

### DIFF
--- a/feature-detects/css/backgroundposition-shorthand.js
+++ b/feature-detects/css/backgroundposition-shorthand.js
@@ -21,12 +21,11 @@ element's background-position simultaniously.
 
 eg `background-position: right 10px bottom 10px`
 */
-define(['Modernizr', 'createElement'], function( Modernizr, createElement ) {
+define(['Modernizr', 'setCss', 'createElement'], function( Modernizr, setCss, createElement ) {
   Modernizr.addTest('bgpositionshorthand', function() {
     var elem = createElement('a');
-    var eStyle = elem.style;
     var val = 'right 10px bottom 10px';
-    eStyle.cssText = 'background-position: ' + val + ';';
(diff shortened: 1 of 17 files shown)


quinncomendant:
What is the status of this fix? I too am trying to find a solution to using Modernizr with CSP.

RehanSaeed:
+1

quinncomendant:
FYI This is being discussed in issue [#1262](https://github.com/Modernizr/Modernizr/issues/1262)

remyabel:
Any movement on this? According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CSSRule/cssText), `cssText`:

> Note: Can no longer be set directly, as now specified to be functionally modify-only -- and *silently* so -- meaning that attempting to set it does absolutely *nothing* -- not even warn or error.  Further, it has no settable sub-properties.  Therefore, to modify, use the stylesheet's cssRules[index] properties .selectorText and .style (or its sub-properties).  See Using dynamic styling information for details.

So this is a pretty important PR in general, if I'm not mistaken.

Identekit:
Is there a fix for this that doesn't involve 'unsafe-eval' or 'unsafe-inline'?  I just downloaded the latest from NuGet, v. 2.8.3 but still received the following errors:

Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' 'nonce-...'". Either the 'unsafe-inline' keyword, a hash ('sha256-CwE3Bg0VYQOIdNAkbB/Btdkhul49qZuwgNCMPgNY5zw='), or a nonce ('nonce-...') is required to enable inline execution.

image

Issue #733, opened 2012-10-19
false positive in history management support detection


bitinn:
This issue has been known for a while but I haven't found a solid way to workaround it.

History API support used by modernizr and majority of routing frameworks does this:

`!!(window.history && history.pushState)`

This is fine on desktop browser, but for mobile browser, in particular android 2.2 - 4.0 and iOS 4, this leads to false positive, where they claim support for pushState, but has issue such as "location.pathname is not updated" or "address bar does not update"

reference:
http://code.google.com/p/android/issues/detail?id=17471
http://caniuse.com/#search=history

This is an issue because we cannot provide a fallback (e.g. hashchange) as it is not reliably detected.

one might say "forget it and let mobile device do a full refresh", but for Web App this is quite critical as full page rendering slow down workflow, even if we assume client-side caching is working.

Hope someone can comment on this.

PR #1270, opened 2014-03-18
History management support feature detection (WIP)

gpoitch:
There have been several discussions on the topic: https://github.com/Modernizr/Modernizr/issues/733, https://github.com/Modernizr/Modernizr/pull/746, https://github.com/Modernizr/Modernizr/pull/891

Certain old browsers are returning a false positive, all leading to just using a UA sniff.
I haven't seen any attempts for an actual feature detect to rule it out.
This is a work in progress. Hoping this thread can spark some ideas.

Attempt 1:
http://jsbin.com/hexifuvi/1/edit
Update state; compare location; put state back.
Works, but touching the page's state is not desired.

Attempt 2:
http://jsbin.com/hexifuvi/5/edit
Run attempt 1 in an iframe.
Works, except in Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=777526)

### DIFF
--- a/feature-detects/history.js
+++ b/feature-detects/history.js
@@ -20,22 +20,15 @@ Detects support for the History API for manipulating the browser session history
 */
 define(['Modernizr'], function( Modernizr ) {
   Modernizr.addTest('history', function() {
+    // The standard check
+    var support = (window.history && 'pushState' in window.history);
+
     // Issue #733
     // The stock browser on Android 2.2 & 2.3, and 4.0.x returns positive on history support
-    // Unfortunately support is really buggy and there is no clean way to detect
-    // these bugs, so we fall back to a user agent sniff :(
-    var ua = navigator.userAgent;
-
-    // We only want Android 2 and 4.0, stock browser, and not Chrome which identifies
-    // itself as 'Mobile Safari' as well
-    if ((ua.indexOf('Android 2.') !== -1 ||
-        (ua.indexOf('Android 4.0') !== -1)) &&
-        ua.indexOf('Mobile Safari') !== -1 &&
-        ua.indexOf('Chrome') === -1) {
-      return false;
+    if(support) {
+      // TODO: WIP
     }
 
-    // Return the regular check
-    return (window.history && 'pushState' in window.history);
+    return support;
   });
 });


gdub22:
@patrickkettner I got the bins resurrected.  Firefox still appears to be an issue as of v34

jokeyrhyme:
I think the wisdom from last I checked was that it's okay (as a last resort) to sniff as part of a blacklist, e.g. these browsers are definitely broken.
What if you had a multi-stage approach:
1. check if the browser is in the blacklist
2. if not, then perform the rest of the feature detect

You might even be able to avoid the iframe test based on the results of the non-iframe test. The downside is that Firefox might end up having to run all parts of the test which could be slow. /shrug

patrickkettner:
yep! thats generally the logic. and that is what is currently in the test.
It would just be great ot avoid the UA sniff totally.

On Mon, Jan 5, 2015 at 5:48 PM, Ron Waldon <notifications@github.com> wrote:

> I think the wisdom from last I checked was that it's okay (as a last
> resort) to sniff as part of a blacklist, e.g. these browsers are definitely
> broken.
> What if you had a multi-stage approach:
> 1. check if the browser is in the blacklist
> 2. if not, then perform the rest of the feature detect
>
> You might even be able to avoid the iframe test based on the results of
> the non-iframe test. The downside is that Firefox might end up having to
> run all parts of the test which could be slow. /shrug
>
> —
> <https://github.com/Modernizr/Modernizr/pull/1270#issuecomment-68796352>.
>



-- 
patrick

feature-detects/history.js
```diff
@@ -20,22 +20,15 @@ Detects support for the History API for manipulating the browser session history
...
-
-    // We only want Android 2 and 4.0, stock browser, and not Chrome which identifies
-    // itself as 'Mobile Safari' as well
-    if ((ua.indexOf('Android 2.') !== -1 ||
-        (ua.indexOf('Android 4.0') !== -1)) &&
-        ua.indexOf('Mobile Safari') !== -1 &&
-        ua.indexOf('Chrome') === -1) {
-      return false;
+    if(support) {
+      // TODO: WIP
```
rejas: @gpoitch is this PR still WIP or do you want to abandon it?

patrickkettner:
sure!

Issue #1204, opened 2014-01-31
Add exports statement as a build option for commonjs


ralphholzmann:
We use an implementation of CommonJS at work. It'd be great if the build page had an option to include an exports statement, so we'd be able to

``` javascript
var Modernizr = require("modernizr");
```

woldie:
I've seen some pretty nice affordances for CommonJS and AMD in libraries like URI.js.  At the top of this module, https://github.com/medialize/URI.js/blob/master/src/URI.js , you'll see a first attempt to export the object in the CommonJS style, then AMD, then finally as a global on the window object if the CommonJS or AMD container was not detected.

A few patterns for implementing this adaptive export can be seen here:  https://gist.github.com/CrocoDillon/9990078

JoshuaWise:
@woldie +1, this would be an excellent upgrade. Instead of automatically adding Modernizr to the global `window`, it would be much better to try to detect the context that the developer is using.

ljharb:
At this point in the world of JS, I don't think anything should be trying to detect the context. I think it would be best if Modernizr strictly was a node-style commonJS module, and those who wish to use it in other contexts can trivially run it through browserify, webpack, or similar. UMDs go out of date, or out of vogue, and add a lot of boilerplate that most people don't need.

JoshuaWise:
@ljharb This would be just is good, imo. I just don't like Modernizr forcing itself into the global context.

SlexAxton:
I think I definitely agree that UMD has not stood the test of time. And I agree that in general commonjs is the default authorship format these days.

But for Modernizr, it can't run outside of a browser. And it can't be loaded nicely with another copy of modernizr (it relies on global browser state, and modifies class names on the html element in most people's builds). It's explicitly a global/singleton dependency.

I'm not against this idea, but I think it's much more than a `s/window/module.exports/g`.

SlexAxton:
I've found in my apps, I generally include a tiny, built modernizr inline, let it leak the global and mess with the `<html>` element, and then when my app spins up, I shim a module over the global:

```js
// AMD
define('modernizr', [], function() {
  return window.Modernizr;
});

// CommonJs
module.exports = window.Modernizr;
```

This provides good performance, by allowing you to load Modernizr in the head of your document, without a request, or the need for a module system, and without thrashing the page render by adding classes asynchronously. And it still allows you to have an explicit dependency tree.

SlexAxton:
[...]

PR #1318, opened 2014-04-25
umd wrapper. solves #1204

jtangelder:
This adds support for commonjs/amdjs. See #1204 for more info.


ryanseddon:
lgtm @patrickkettner anything else you were waiting on?

patrickkettner:
I always prefer 1 commit, but otherwise :+1:

ryanseddon:
@jtangelder can you squash to 1 commit and then we'll merge.

stucox:
Could you add a unit test? Slightly scared the builder, or some other change, *could* break this.

Plus we should get hotter on demanding tests for system changes.

stucox:
FYI I don't think we'll have time to document this thoroghly for v3, so either it'll be an undocumented feature until 3.1, or we should hold off on merging.

patrickkettner:
I'm fine with holding off until 3.1

PR #1338, opened 2014-05-09
Always refer to window on the htmlshiv

jtangelder:
I got in a situation wih loading modernizr within an eval where the this scope was undefined at the htmlshiv. Changing this to window is a save fix since it it given as an argument. It even is shorter when minified.


patrickkettner:
closing as outdated

PR #1352, opened 2014-05-20
Doc block updates

stucox:
**Don’t merge yet.**

I’ve been going alphabetically; current progress: `feature-detects/css/h*`

If anyone fancies helping, they can raise a PR against [the `docblocks` branch](https://github.com/Modernizr/Modernizr/tree/docblocks) (or core team can just push straight to that branch).

One thing I haven’t done is cross-referenced this against the [docs page on the site](http://modernizr.com/docs/#s2) – I think ultimately we want to move all of those examples into the relevant doc blocks, then populate the v3 site from metadata.

Not considering this a v3 blocker… more a continual work in progress.


patrickkettner:
@stucox any chance you would be willing to get this in a mergeable state?

stucox:
The only reason I said *“Don’t merge yet”* is so we could get a load in together in a nice clean single commit. I’ll bring this up to date with `master` then we can merge it as-is, and tackle any that missing doc blocks later.

patrickkettner:
Thanks stu

PR #1428, opened 2014-09-24
Fix: oninput detection for IE9-, csshyphens detection for FF and Safari

anenviousguest:
# `oninput` detection fix for IE9-

The oninput event detection was not as precise in IE9- as it could be. The oninput test was reporting `undefined` instead of `false`, thus making unit tests failing:
- 5. Modernizr classes and bools: html classes are looking good<br>
  Test number **163**: oninput is correctly false in the classes and objectExpected:  false Result:  undefined Diff:  false undefined  
- 6. Modernizr classes and bools: Modernizr properties are looking good<br>
  Test number **283**: Modernizr.oninput is a boolean value or Boolean object

This was caused by fact that events are created and fired differently in IE9-, see respective code change.
# `csshyphens` detection fix for FF and Safari
- There was seemingly a typo in CSS property name (`text-justification` instead of correct `text-justify`).
- The hyphenation doesn't take effect unless the `lang` attribute is explicitly set, for this reason it was erroneously detected as not supported on FF and Safari, whereas it **is** actually supported.

### DIFF
--- a/feature-detects/css/hyphens.js
+++ b/feature-detects/css/hyphens.js
@@ -60,20 +60,21 @@ define(['Modernizr', 'prefixes', 'createElement', 'testAllProps', 'addTest'], fu
           var result = false;
           var firstChild = document.body.firstElementChild || document.body.firstChild;
 
+          // Hyphenation is only applied when language is explicitly set and when respective dictionary
+          // is available. See https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens for details.
+          div.lang = 'en';
           div.appendChild(span);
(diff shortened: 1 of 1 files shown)


anenviousguest:
Bumping the thread. Any chance to land this?
Thanks!

rejas:
hi @anenviousguest sorry for the long delay. if you are still interested, would you mind splittin this PR up and submit only the IE9 stuff? Hypens dectection should go into a seperate PR (the justification already is in a PR, so only the lang part would be needed) Thx

anenviousguest:
hi @rejas, will do, thanks

rejas:
so, merged your other PR #2345 
this PR needs to get rebased and updated to only include the ie9 stuff if it needs to go foward @anenviousguest

codeCraft-Ritik (commented):
Nice fixes! Improving event detection in IE9- and correcting CSS property issues makes the tests more reliable.
Good catch on the hyphenation behavior and lang requirement

PR #1429, opened 2014-09-24
Add option to toggle no-js replacement

### LINKED ISSUE #1385: Option to stop Modernizr converting "no-js" to "js"
I would like to propose a new option to govern this default behaviour.  Perhaps:

```
ModernizrProto._config['jsSupportClass'] = true | false | ["from", "to"];
```

If true (default behaviour for backwards compatibility) it replaces `no-js` with `js`.  If false, it doesn't do anything.  If an array, it replaces the first value with the second.

Use case: A site (eg ft.com) chooses to set the bar for JavaScript execution higher than the simple presence of some kind of JavaScript engine.  This approach has been popularised by the BBC and was named [Cuts the mustard](http://responsivenews.co.uk/post/18948466399/cutting-the-mustard) by their dev team.  UAs that do not 'cut the mustard' should be treated as if they have no JS capability, thereby reducing testing overhead and easing support of legacy clients.

This will increasingly become more of an issue as the performance gap between ES6-supporting UAs and those that support only ES3 challenges developers to write increasingly bizarre code to take advantage of ES6 features while somehow continuing to provide some capability for legacy clients.

We also wish to support a more complex use case in which there are multiple cuts the mustard tests - eg one for dynamic advertising and analytics, and a higher one for everything else.

If there's agreement from maintainers on an approach I'm happy to code it up and submit a PR.

Cheers.

rose:
This resolves #1385.

### DIFF
--- a/src/ModernizrProto.js
+++ b/src/ModernizrProto.js
@@ -8,6 +8,7 @@ define(['tests'], function ( tests ) {
     _config: {
       classPrefix : '',
       enableClasses : true,
+      enableJSClass : true,
       usePrefixes : true
     },
 
--- a/src/setClasses.js
+++ b/src/setClasses.js
@@ -5,11 +5,13 @@ define(['Modernizr', 'docElement'], function( Modernizr, docElement ) {
     var className = docElement.className;
     var classPrefix = Modernizr._config.classPrefix || '';
 
-    // Change `no-js` to `js` (we do this regardles of the `enableClasses`
+    // Change `no-js` to `js` (we do this independently of the `enableClasses`
     // option)
     // Handle classPrefix on this too
-    var reJS = new RegExp('(^|\\s)'+classPrefix+'no-js(\\s|$)');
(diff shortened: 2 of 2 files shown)


rose:
Wow, I totally dropped this.  It seems like there should be a way to test config options, but my js-fu is not sufficient to determine the right way to do it.  Should I open a separate issue for that?

rose:
In fact, looking at it again I'm pretty sure there's a bug there.  

In the old version of setClasses it says that no-js gets replaced regardless of whether enableClasses is set.  But the only assignment to docElement.className happens inside the if statement for enableClasses.  Either the comment or the code is wrong.

patrickkettner:
@rose for the moment, the "real" work is getting proper coverage reporting into the repo, at least in my view. I think there is a fair amount of stuff that should be tested in `ModernizrProto` that isn't currently, and the testing for this could easily be made much easier once we figure out how to do that. I just would like to get a lay of the land first.

regarding the bug, thats a comment issue, rather than code

PR #1443, opened 2014-10-10
Fix JS error on Firefox hidden iframes: verify that getComputedStyle is not returning null before using it

felipebrahm:
On Firefox, `getComputedStyle` returns `null` if it's called inside a hidden iframe. You can see an example here:
https://www.groupon.com/deals/butter

Open that URL on Firefox. You'll see several JS errors in the console, the first being `TypeError: window.getComputedStyle(...) is null`. This is happening not on that page, but on a hidden iframe on that page (that iframe is loaded on page load but it's only shown when a user takes certain action, so it's a completely valid use case).

So, basically, this code will break:
```javascript
var result = false;
if(window.getComputedStyle) {
  result = window.getComputedStyle(element, null).getPropertyValue('font-size') === '10px';
}
return result;
```

The solution is to first check for the returned value, like this:
```javascript
var result;
var style;
if(window.getComputedStyle && (style = window.getComputedStyle(element, null)) !== null) {
  result = style.getPropertyValue('font-size') === '10px';
}
return result;
```

Different files have different coding styles, so I tried to preserve the coding style for each file.

### DIFF
--- a/feature-detects/css/pseudoanimations.js
+++ b/feature-detects/css/pseudoanimations.js
@@ -8,6 +8,7 @@
 define(['Modernizr', 'testStyles', 'test/css/animations'], function (Modernizr) {
   Modernizr.addTest('csspseudoanimations', function () {
     var result = false;
+    var style;
 
     if (!Modernizr.cssanimations || !window.getComputedStyle) {
       return result;
@@ -21,7 +22,10 @@ define(['Modernizr', 'testStyles', 'test/css/animations'], function (Modernizr) 
     ].join('');
 
     Modernizr.testStyles(styles, function (elem) {
-      result = window.getComputedStyle(elem, ':before').getPropertyValue('font-size') === '10px';
+      style = window.getComputedStyle(elem, ':before');
+      if(style !== null) {
+        result = style.getPropertyValue('font-size') === '10px';
+      }
     });
 
     return result;
--- a/feature-detects/css/pseudotransitions.js
+++ b/feature-detects/css/pseudotransitions.js
@@ -8,6 +8,7 @@
 define(['Modernizr', 'testStyles', 'test/css/transitions'], function (Modernizr) {
   Modernizr.addTest('csspseudotransitions', function () {
     var result = false;
+    var style;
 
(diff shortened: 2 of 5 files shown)


benrbray:
Has any progress been made here?  I'm having the same problem where Modernizr completely fails in Firefox.

dmethvin:
I agree with @patrickkettner that adding all those null checks is going to be error-prone for anyone writing a Modernizr detect. Abstracting this into a function would help to centralize the noise.

Looking at [the current uses](https://github.com/Modernizr/Modernizr/search?utf8=%E2%9C%93&q=getcomputedstyle), the only complicated case is [the one for ruby](https://github.com/Modernizr/Modernizr/blob/347ddb078116cee91b25b6e897e211b023f9dcb4/feature-detects/elem/ruby.js#L43-53) which already has its own wrapper. Most of the others seem to just need one property off `getComputedStyle()` rather than a generic callback. So something like:
```js
function getStyleProperty( style, pseudo, prop ) {
   // IE <= 8 doesn't have getComputedStyle
   // Firefox 39+ returns null for styles in hidden iframes
   var cs = window.getComputedStyle && window.getComputedStyle( style, pseudo );
   if ( !cs ) {
      return null;
   }
   return cs.getPropertyValue( prop );
}
```
There are some differences in usage like [vhunit](https://github.com/Modernizr/Modernizr/blob/347ddb078116cee91b25b6e897e211b023f9dcb4/feature-detects/css/vhunit.js#L20-22) and [inputtypes](https://github.com/Modernizr/Modernizr/blob/347ddb078116cee91b25b6e897e211b023f9dcb4/feature-detects/inputtypes.js#L80) but those should work with `.getPropertyValue()` too.

When `.getComputedStyles()` doesn't work it seems really likely these detects will just be broken and maybe give the wrong results. If the iframe is permanently hidden I doubt that matters, but for cases where it starts hidden and is later shown that could cause problems.

patrickkettner:
thanks for the initial repotr @felipebrahm! With the new helper and associated eslint rule, this shouldn't ever be a problem again

PR #1463, opened 2014-11-27
Attempt to use replaceState to validate we can use the History API.

stevemckenzie:
Checking if history (pushState) is supported will pass for Firefox Addons with the current implementation but is actually not supported. You cannot tell though until you attempt to actually use the API.

For more info, I originally reported this here: https://github.com/rackt/react-router/issues/535

Also note the `TODO`. Looking for suggestions for that.

### DIFF
--- a/feature-detects/history.js
+++ b/feature-detects/history.js
@@ -35,6 +35,15 @@ define(['Modernizr'], function( Modernizr ) {
       return false;
     }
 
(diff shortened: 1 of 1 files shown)


ryanseddon:
@stevemckenzie, @patrickkettner  bump what do we need to do to get this merged?

PR #1469, opened 2014-12-07
fixes for the flash detect

### LINKED ISSUE #1413: Modernizr v3 flash detect always returning {block: true} in Chrome
Hey,

When using the Flash feature detection in Modernizr v3, `Modernizr.flash` always has the value `{blocked: true}` when using Chrome (or equivalent like Opera). 

Flash is correctly enabled in the browser and the page in question includes Modernizr in the `<head>`. If the same page is tested in Firefox, `Modernizr.flash` has the value `true`.

The code that is causing the issue appears to be https://github.com/Modernizr/Modernizr/blob/master/feature-detects/flash.js#L54. `<embed>` does not possess the property `Pan` at the point that Modernizr performs its check. 

Shifting the code inside the `setTimeout` does not help, nor does increasing the timeout. However, if you execute the following code in the console after the page has fully loaded it will return `true`

```javascript
var embed = document.createElement('embed');
embed.type = 'application/x-shockwave-flash';
document.body.appendChild(embed);
console.log('Pan?', ('Pan' in embed));
```

I'm using Chrome v36, Mac OSX.

### LINKED ISSUE #1432: flash detection throws an error in IE
The feature-detect/flash.js script throws an error in all version of IE.

this happens at lin 82 :
if (body.fake) {
    body.parentNode.removeChild(body);
}

the error is : SCRIPT5007: Unable to get property 'removeChild' of undefined or null reference
this happens because the parentNode property is empty.

i've 'fixed' this in my local version like this : 
if (body.fake && body.parentNode) {
    body.parentNode.removeChild(body);
}

Although i think this is not the correct way to actually fix it...
just wanted to report the error and see if anybody knows how to correctly fix this issue.

ps: this is on the master branch

patrickkettner:
fixes #1413, also fixes #1432

### DIFF
--- a/feature-detects/flash.js
+++ b/feature-detects/flash.js
@@ -19,7 +19,7 @@ define(['Modernizr', 'createElement', 'docElement', 'addTest', 'getBody'], funct
         bool.blocked = (result === 'blocked');
       }
       addTest('flash', function() { return bool; });
-      if (embed) {
+      if (embed && body.contains(embed)) {
         body.removeChild(embed);
       }
     };
@@ -42,12 +42,14 @@ define(['Modernizr', 'createElement', 'docElement', 'addTest', 'getBody'], funct
       // actually create an element to see what happens to it.
       var embed = createElement('embed');
       var body = getBody();
+      var blockedDetect;
       var inline_style;
 
       embed.type = 'application/x-shockwave-flash';
 
       // Need to do this in the body (fake or otherwise) otherwise IE8 complains
       body.appendChild(embed);
+      docElement.appendChild(body);
 
       // Pan doesn't exist in the embed if its IE (its on the ActiveXObjeect)
       // so this check is for all other browsers.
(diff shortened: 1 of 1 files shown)


feature-detects/flash.js
```diff
@@ -77,11 +85,18 @@ define(['Modernizr', 'createElement', 'docElement', 'addTest', 'getBody'], funct
           }
         }
 
-        // If we’re rockin’ a fake body, clean it up
-        if (body.fake) {
+        // If we’re rockin’ an attached fake body, clean it up
+        if (body.fake && body.parentNode) {
```
noahgrant: Not sure if it stems from this, and I can create an issue for it, but my Modernizr build from gulp-modernizr (after a fresh npm install yesterday) did not remove the extra body element, so we had two `<body>`s in our app!  I had to set modernizr to "3.0.0-pre" in our package.json, since customizr uses the master branch from github.

patrickkettner: would you be able to test this commit and the commit before?

noahgrant: Confirmed—installing Modernizr at commit [https://github.com/Modernizr/Modernizr/commit/a6ca9d9f240071107249241f83223c4185151875](a6ca9d) works as normal, but Modernizr at commit [https://github.com/Modernizr/Modernizr/commit/1db6eee599e40848b5f571f3c09f32b64b926d82](1db6ee) leaves the extra body element in.

patrickkettner: what browser? are you loading in the head body, or bottom?

noahgrant: happens in chrome 39, firefox 32, and safari 8, with the script in the head.

niksy: Seeing the same thing, so I created [pull request](https://github.com/Modernizr/Modernizr/pull/1506) for this.

PR #1481, opened 2015-01-07
Add bang so minification is easier

garrypolley:
Depending on the minification that's used having the `/*! ... comment ..*/` is helpful.  I suggest adding the bang here so the custom builds do not have to modify the source that's generated and add a `!` to ensure they maintain the MIT copyright when minifying. 

I think it's the custom build that reads from this file to add the license.  I could be wrong, in which case, I hope the thing that does the custom build adds the `!` to the start of the copyright/license comment.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,1 +1,1 @@
-/* Modernizr 3.0.0pre (Custom Build) | MIT  */
+/*! Modernizr 3.0.0pre (Custom Build) | MIT  */


patrickkettner:
Hey @garrypolley!

> Depending on the minification that's used

Which situations are that?

garrypolley:
@patrickkettner Thanks for responding. 

Specifically we are using the uglifier minifier with a Rails application. When we make a custom build of Modernizr it does not include the `/*!` at the start of the license comment.  We'd like it to have that (or the word copyright) in the body of the comment so we can keep the license more easily in our minified code.  

It may not seem like a big deal, but it's important to us to attribute all the awesome libraries (like this one) that we use in our production code.  We'd prefer not to have to modify the source to keep the license comment.

patrickkettner:
Nah, totally cool with adding the `!`, just wanted the reasons more explicit

PR #1495, opened 2015-01-20
Improves error handling in bind, per es5-shim implementation

rgkimball:
The Function.prototype.bind, borrowed from es5-shim, is using typeof to determine whether arguments passed through bind are functions or not. This has the unintended side effect of inadvertently halting js processing in older browsers, where a try/catch would be more appropriate.

A few others have experienced this problem but no solution is proposed:
http://stackoverflow.com/questions/12569030/modernizr-causing-errors-in-ie8-exception-handling
https://github.com/pimterry/loglevel/issues/21
http://w3facility.org/question/modernizr-causing-errors-in-ie8-exception-handling/
https://www.bountysource.com/issues/1569186-stack-overflow-error-when-running-a-simple-app-in-ie8-started-on-meteor-0-8-0

The es5-shim library from which this functionality originates appropriately uses an isFunction helper to check whether the object is a function instead. The following commit introduces this check to the modernizr adaptation.

https://github.com/es-shims/es5-shim/blob/master/es5-shim.js#L56

### DIFF
--- a/modernizr.js
+++ b/modernizr.js
@@ -246,8 +246,11 @@ window.Modernizr = (function( window, document, undefined ) {
 
         var target = this;
 
-        if (typeof target != "function") {
-            throw new TypeError();
+        try {
+          Object.prototype.toString.call(target) === '[object Function]';
+        }
+        catch(err) {
+          throw new TypeError(err);
         }
 
         var args = slice.call(arguments, 1),


rgkimball:
Noting that the prior commit for 2.8.3 also fails the travis build:

https://travis-ci.org/Modernizr/Modernizr/builds/29075521

stucox:
@patrickkettner Is this something we should include in `backport-fixes`? Or should we consider whether to apply it to v3 (if such equivalent code still exists)?

stucox:
*Aside:* Just checked the code and we don’t have a `.bind` polyfill in v3 anymore – we just use this function instead: https://github.com/Modernizr/Modernizr/blob/master/src/fnBind.js

patrickkettner:
if we are backporting something, I would rather use the `fnBind` that is used in master

stucox:
That could be a breaking change, as some v2 users might depend on the .bind polyfill (which is effectively “leaked”) — even though it isn't documented.

So I'd rather be cautious and leave this, and save backport fixes for critical bugs.

Sorry @rgkimball, thanks for the PR though.

stucox:
Happy to reopen if you can demonstrate a use case which is functionally broken by the `typeof` check.

PR #1503, opened 2015-01-27
Use `window` instead of `this` to set scoped global object reference

### LINKED ISSUE #1431: Modernizer w/ CommonJS (Browserify)
Due to the fact that Modernizr uses `this` as a reference to `window`, wrapping it in a closure breaks Modernizr.  

Is there a reason I don't know about for using `this` instead of window?  Because it certainly seems like this is a fickle practice with modern build systems.

redoPop:
This pull request resolves #1431. Per my comment on that issue:

> Modernizr's assumption that `this === window` is causing it to break for an increasing number of us who use Browserify and similar tools. Wherever Modernizr's code is executed with a non-global function context, it will break. We can work around it by adding some extra steps to our build tools, but it could be avoided entirely by replacing the offending line with something like `})(window, document);`
>
> `this` was once used to shorten walks up the scope, [but `window` seems to work just as well for that](http://jsperf.com/this-vs-window-for-global-object-access).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -141,7 +141,7 @@ module.exports = function( grunt ) {
           fileExclusionRegExp: /^(.git|node_modules|modulizr|media|test)$/,
           wrap: {
             start: '<%= banner.full %>' + '\n;(function(window, document, undefined){',
-            end: '})(this, document);'
+            end: '})(window, document);'
           },
           onBuildWrite: function (id, path, contents) {
             if ((/define\(.*?\{/).test(contents)) {


paulirish:
I got no qualms with that.

patrickkettner:
This was changed with the larger refactor, but the intention of the change was merged. Thank you very much, @redoPop for bringing it up!

jeromecovington:
I see this was backported to 2.x but when I hit this CDN I still see:
```
})(this, this.document);
```

https://cdnjs.cloudflare.com/ajax/libs/modernizr/2.8.3/modernizr.js

redoPop:
Looks like the PR for the backport is still open: #1608

PR #1504, opened 2015-01-27
update webgl test to use probablySupportsContext

### LINKED ISSUE #1499: WebGL false positive in Safari 7
tested in Safari 7.1.2 (OS X 10.9.5)

If WebGL has not been enabled in the Develop menu, `WebGLRenderingContext` is still found on `window` and `supportsContext` was not added to later builds of Safari 7 for some reason, so the current test returns true.

I haven't found a workaround yet, short of the old, slow way of actually trying (and failing) to create a WebGL context.

brendankenny:
fixes #1499 

spec was updated to use `probablySupportsContext` instead of `supportsContext`: https://developers.whatwg.org/the-canvas-element.html#dom-canvas-probablysupportscontext

WHATWG thread: https://lists.w3.org/Archives/Public/public-whatwg-archive/2013Sep/0013.html

updated in WebKit here: https://bugs.webkit.org/show_bug.cgi?id=120716

AFAICT `supportsContext` was only ever implemented in WebKit and `supportsContext` was replaced with `probablySupportsContext` in WebKit three months later, so it seems sufficient to just replace `supportsContext`. I've only tested in Safari 7 and iOS 7, however.

### DIFF
--- a/feature-detects/webgl.js
+++ b/feature-detects/webgl.js
@@ -10,9 +10,10 @@
 define(['Modernizr', 'createElement'], function( Modernizr, createElement ) {
   Modernizr.addTest('webgl', function() {
     var canvas = createElement('canvas');
-    if ('supportsContext' in canvas) {
-      return canvas.supportsContext('webgl') || canvas.supportsContext('experimental-webgl');
+    var supports = 'probablySupportsContext' in canvas ? 'probablySupportsContext' :  'supportsContext';
+    if (supports in canvas) {
+      return canvas[supports]('webgl') || canvas[supports]('experimental-webgl');
     }
-    return !!window.WebGLRenderingContext;
+    return 'WebGLRenderingContext' in window;
   });
 });


patrickkettner:
thanks so much for the fast PR, @brendankenny!

image


looks like we will need to support both

brendankenny:
Ah, bizarre. Updated to test both. PTAL

patrickkettner:
beautiful squishing :]

feel like we could dry it up a bit, though. maybe something like

```
var api = 'supportsContext' in canvas ? 'supportsContext' :  'probablySupportsContext'; 

if (api in canvas) {
    return canvas[api]('webgl') || canvas[api]('experimental-webgl');
}
```

brendankenny:
Updated.

That second conditional really bugs me, but I haven't figured out a simpler way that gets around it :)

feature-detects/webgl.js
```diff
@@ -10,9 +10,10 @@
...
   Modernizr.addTest('webgl', function() {
     var canvas = createElement('canvas');
-    if ('supportsContext' in canvas) {
-      return canvas.supportsContext('webgl') || canvas.supportsContext('experimental-webgl');
+    var supports = 'probablySupportsContext' in canvas ? 'probablySupportsContext' :  'supportsContext';
+    if (supports in canvas) {
+      return canvas[supports]('webgl') || canvas[supports]('experimental-webgl');
     }
-    return !!window.WebGLRenderingContext;
+    return 'WebGLRenderingContext' in window;
```
patrickkettner: awesome catch!

patrickkettner:
the only way around it I can think of would be something like
```javascript
var supports;
if (supports = 'probablySupportsContext' in canvas ? 'probablySupportsContext' :  'supportsContext' in canvas ? 'supportsContext' : false) {
  //...
}
```
but that feels ugly

patrickkettner:
actually, you could avoid the ternary all together if you wanted. was in a everything-is-a-nail, mode

```
var supports;
if (supports = (canvas.probablySupportsContext || canvas.supportsContext || false)) {
  return supports('webGl')
}
```

brendankenny:
yeah, I had something like that, but you need to call it on the canvas element, so you end up needing something like
```
var supports = canvas.probablySupportsContext || canvas.supportsContext;
if (supports) {
  return supports.call(canvas, 'webgl') || supports.call(canvas, 'experimental-webgl');
}
```
I asked a couple of people and they liked your version better :)

patrickkettner:
okey doke!
so we are all happy with the current setup?

brendankenny:
this looks good to me

patrickkettner:
@brendankenny thanks for all the work!

cheers!

PR #1506, opened 2015-01-30
Remove fake body for truthy in all non-IE browsers

niksy:
Flash detect doesn’t remove fake body when it passes `Pan` test in non-IE browsers, and since check returns early, it won’t get to `blockedDetect` check and it won’t remove fake body.

@patrickkettner, since you’re the who created the detect, could you please review it and see if this makes any sense?


patrickkettner:
beautiful work! so sorry it took so long to review

PR #1510, opened 2015-02-04
Update QUnit to 1.17.1

leobalter:
Running the tests for another issues, I saw that updating QUnit to 1.17.1 makes no harm to the test suite. This version also prevents any async leakage. that should be good.


patrickkettner:
thanks so much for opening up a PR, @leobalter, but this is going to be superseded by #1531 

look forward to more contributions from you!

cheers

Issue #15, opened 2009-10-17
3d transforms false positive in chrome


paulirish:
http://www.deepbluesky.com/blog/-/browser-support-for-css3-and-html5_72/#comment_145

PR #1513, opened 2015-02-05
csstransform3d detect should prefer css conditionals - fixes #1512

### LINKED ISSUE #1512: IE/Spartan EdgeHTML engine fails CSSTransforms3d detection
Using Windows 10 Tech Preview (or http://remote.modern.ie), browse to http://modernizr.github.io/Modernizr/test/

Expect: csstransforms3d: true
Actual: csstransforms3d: false

http://caniuse.com/#feat=transforms3d 

(Note: we haven't flipped the new [EdgeHTML engine](http://blogs.msdn.com/b/ie/archive/2014/11/11/living-on-the-edge-our-next-step-in-interoperability.aspx) on by default yet but will before we ship.  To do enable it for this issue, go to about:flags and enable Experimental Web Platform Features)

jacobrossi:
This fixes https://github.com/Modernizr/Modernizr/issues/1512
There's several ways this could be done.  This looked cleanest to me.

The root cause is that EdgeHTML looks a lot like webkit now.  So it falls through the existing double-check for a old webkit bug (referenced in the code comments, also see  https://github.com/Modernizr/Modernizr/issues/15) where css 3d transforms were sometimes not correctly detected due to GPU acceleration.

However, that double-check relies on the non-standard [at]media(transform-3d) media query.  We've been discussing this with the W3C and consensus is that this media query should go away from browsers, but that's contingent on updating content relying on it (including Modernizr).

Context: http://lists.w3.org/Archives/Public/www-style/2014Oct/0390.html and https://code.google.com/p/chromium/issues/detail?id=426644 

So my proposed fix here is to prefer the standardized CSS supports detection mechanism and only fall back to the [at]media rule when CSS at supports isn't available.

I've tested and validated this on IE (6-11 doc modes), Spartan, Chrome, Firefox, Opera (blink-based), and Safari (Mac and iOS).

Note that this fix won't work in the latest public EdgeHTML preview builds quite yet still because it doesn't have [at]supports. But I have special builds. :-) Coming soon to public preview builds.

### DIFF
--- a/feature-detects/css/transforms3d.js
+++ b/feature-detects/css/transforms3d.js
@@ -9,7 +9,7 @@
   ]
 }
 !*/
-define(['Modernizr', 'testAllProps', 'testStyles', 'docElement'], function( Modernizr, testAllProps, testStyles, docElement ) {
+define(['Modernizr', 'testAllProps', 'testStyles', 'docElement', 'test/css/supports'], function( Modernizr, testAllProps, testStyles, docElement ) {
   Modernizr.addTest('csstransforms3d', function() {
     var ret = !!testAllProps('perspective', '1px', true);
     var usePrefix = Modernizr._config.usePrefixes;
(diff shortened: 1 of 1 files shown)


patrickkettner:
looks awesome, @jacobrossi!

only issue issue is we should be importing the existing supports check (add `test/css/supports` to the require statement), and then key off of `Modernizr.supports`, rather than having a separate detect.

jacobrossi:
Ah yes, that makes sense.  Done!

patrickkettner:
@jacobrossi any chance you could squish down the commits to just one?

cheers!

jacobrossi:
Yep, no problem.

patrickkettner:
beautiful work, @jacobrossi

PR #1514, opened 2015-02-05
Catch window.find exception in FF for Android, fixes #1438

### LINKED ISSUE #1438: CSS hyphens detect broken in Firefox for Android
The CSS hyphens detect leaves an input box and some text on top of the document. These elements get removed from the DOM in other browsers but not in Firefox for Android. The input box also gets focused which brings up the virtual keyboard.

The following screenshot was taken on a Google Nexus 10 (Android 4.4.4) in Firefox 32.0.3.

screenshot_2014-10-08-13-22-30

Here's the JSFiddle that I used: http://jsfiddle.net/d85h6w7w/

ryanhughes:
This prevents lebowskilebowski being left at the top of the page when using Firefox on Android.

Firefox on Android throws a NS_ERROR_NOT_AVAILABLE exception when calling window.find()

Happens on FF35.0 and Android 4.4.4

Screenshot: #1438

### DIFF
--- a/feature-detects/css/hyphens.js
+++ b/feature-detects/css/hyphens.js
@@ -165,15 +165,15 @@ define(['Modernizr', 'prefixes', 'createElement', 'testAllProps', 'addTest'], fu
           }
 
           /* try to find the doubled testword, without the delimiter */
-          if (window.find) {
-            result = window.find(testword + testword);
-          } else {
-            try {
+          try {
+            if (window.find) {
+              result = window.find(testword + testword);
+            } else {
               textrange = window.self.document.body.createTextRange();
               result = textrange.findText(testword + testword);
-            } catch (e) {
-              result = false;
             }
+          } catch (e) {
+            result = false;
           }
 
           document.body.removeChild(div);


myfonj:
Great to see this bug fixed!
There seem to remain some outdated versions of Modernizr scattered around interwebs, but it is nothing you can fix, naturally.

Just one thing would be nice in general in such situation: **if you are making some workaround for browser bug, please, make sure the bug was documented and properly reported**. And if wasn't, please report it.

It seems this fix overcomes problem that exists in Firefox _for years_ (yes, it is still not fixed), yet I wasn't able to find any trace of report narrowing cause of it from times this hilarious "lebowskilebowskigate" emerged, although it seems it was well known to you (developers) and certainly to author of this fix quite soon it exhibited itself.  If it had been reported then, perhaps it would be fixed already and there would not be any manifestation of it at pages with outdated Modernizr today.

Coincidentally, first report matching this bug I could find is from _yesterday_. Leaving here for reference:
https://bugzilla.mozilla.org/show_bug.cgi?id=1654683

PR #1517, opened 2015-02-09
switch to universal requirejs based build system

patrickkettner:
well over a year after @robwierzbowski spearheaded the concept, I finally
finished up a version that works in the browser, node, and grunt.

browser usage would just involve including the `lib/build.js` file, and
then `require`ing the `build` module, like so 

```html
<script src="scripts/modernizr-git/lib/r.js"></script>
<script src="scripts/modernizr-git/lib/build.js"></script>
<script>
var builder;

requirejs(['build'], function(_builder) {
   builder = _builder;
});

$('document').on('buildChangeEventOrSomething', function(config) {
    builder(config, function(output) {
      $('#output').text(output);
    });
});
</script>
```

I'll update the v3 repo once this is merged.

closes #1184


lib/build.js
```diff
@@ -1,86 +1,123 @@
...
+
+var requireConfig = {
+  optimize: 'none',
+  optimizeCss: 'none',
+  useStrict: true,
+  include: ['modernizr-init'],
+  fileExclusionRegExp: /^(.git|node_modules|modulizr|media|test)$/,
+  wrap: {
+    start: '\n;(function(window, document, undefined){',
+    end: '})(this, document);'
```
SlexAxton: There were some requests that we switch our outer wrapper to just use `window` since it absolutely requires a real `window`.

lib/build.js
```diff
@@ -1,86 +1,123 @@
...
+  },
+  onBuildWrite: function (id, path, contents) {
+    if (this.optimize == 'uglify2') {
```
SlexAxton: double equals! The HORROR. (i'm fine with it)

src/generate-banner.js
```diff
@@ -0,0 +1,31 @@
...
+  return function banners(type) {
+    if (!type || type === 'compact') {
+      return '/*! ' + pkg.name + ' ' + pkg.version + ' (Custom Build) | ' + pkg.license  + ' */\n';
+    }
+    else if (type === 'full') {
+      return '/*!\n' +
+        ' * ' + pkg.name + ' v' + pkg.version + '\n' +
+        ' * modernizr.com\n' +
+        ' *\n' +
+        ' * Copyright (c)\n *  ' + _.pluck(pkg.contributors, 'name').join('\n *  ') + '\n\n' +
```
SlexAxton: Should this just be `Copyright Modernizr Team & Contributors`

SlexAxton: (I can just see it getting long)

SlexAxton:
This looks pretty solid. I'd say merge-at-will. Would probably be good to generate a few tests into a file and put it in a gist and see if we can't identify anything else we could be optimizing (or that we're breaking). But this is fantastic.

lib/build.js
patrickkettner: I just ripped this from another part, but yeah, good a time as any

src/generate-banner.js
patrickkettner: eh - its only in the debug build. I'm fine with having eternal glory on poorly optimized sites.

SlexAxton: ha. well said.

patrickkettner:
thanks alex!

cc @doctyper, since this is probably breaking grunt-modernizr (but should ultimately be cleaner)

stucox:
:thumbsup: Great stuff!

Issue #1518, opened 2015-02-09
matchMedia check


matthew-andrews:
Would it be possible to have a test for `matchMedia`?  Thanks!

patrickkettner:
the lovely @AlbertoElias stepped up and took care of it in #1520

matthew-andrews:
mega!

PR #1520, opened 2015-02-10
Adds feature detect and polyfill for window.matchMedia()

AlbertoElias:
Addresses #1518

### DIFF
--- a/feature-detects/window/matchmedia.js
+++ b/feature-detects/window/matchmedia.js
@@ -1,0 +1,25 @@
+/*!
+{
+  "name": "matchMedia",
+  "property": "matchmedia",
+  "caniuse" : "matchmedia",
+  "tags": ["matchmedia"],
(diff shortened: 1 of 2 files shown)


patrickkettner:
thanks for the PR @AlbertoElias!

there is also `msMatchMedia`, could you throw a `prefixed` check around it to cover that?

```
!!Modernizr.prefixed('matchMedia', window);
```

AlbertoElias:
@patrickkettner Thanks for the quick response. Do you mean something like my last commit? Couldn't find the docs for that but found it being used elsewhere.

patrickkettner:
Yep! Thanks so much. Could you squish them down to a single commi and then we will get it in?

AlbertoElias:
@patrickkettner Done :)

patrickkettner:
thanks mang!

PR #1521, opened 2015-02-10
API bug fixes for #1517

doctyper:
Small patch to fix API/CLI issues introduced in Modernizr/Modernizr#1517

### DIFF
--- a/feature-detects/ambientlight.js
+++ b/feature-detects/ambientlight.js
@@ -12,5 +12,5 @@
 Detects support for the API that provides information about the ambient light levels, as detected by the device's light detector, in terms of lux units.
 */
 define(['Modernizr', 'hasEvent'], function( Modernizr, hasEvent ) {
-  Modernizr.addTest('devicelight', hasEvent('devicelight', window));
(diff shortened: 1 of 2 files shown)


feature-detects/ambientlight.js
```diff
@@ -12,5 +12,5 @@
 Detects support for the API that provides information about the ambient light levels, as detected by the device's light detector, in terms of lux units.
 */
 define(['Modernizr', 'hasEvent'], function( Modernizr, hasEvent ) {
-  Modernizr.addTest('devicelight', hasEvent('devicelight', window));
```
doctyper: My Customizr tool fails testing due to this unreferenced test name. I could instead change the file name and leave the test name as it was. Open to either.

patrickkettner: `ambientlight` makes more sense to me

patrickkettner:
great call on the package.json!

how is the cli dependant on `file`?

doctyper:
It's used by the `metadata` endpoint [here](https://github.com/Modernizr/Modernizr/blob/master/lib/metadata.js#L2).

patrickkettner:
doi - don't know how I missed that.

cheers

Issue #1505, opened 2015-01-29
Add a test for video "Javascript intiated" play


jraoult:
Modernizr 3 already has a test to detect if `autoplay` value is fulfilled by the browser. There is another cas that diverges a little bit when the video playback is triggered by JS (method `play()` of the `HTMLMediaElement`).

For example, FF for Android (as of v35) prevents video from playing even if `autoplay` attribute is set to `true` (hence the Modernizr's "videoautoplay" test is negative) but doesn't prevent the video playback to be initiated by JS even without user intervention.

I wrote a Modernizr test for that very similar to "videoautoplay" where instead of setting `autoplay` to `true` I called the method `play()` on the video element.

patrickkettner:
Hey Jonathan!
Thanks a ton for the heads up!
Would you be interested in writing a PR adding that detect? The only main
pointer I would have is to abstract out the gigantic base64 srcs into
another file, then `require`ing them as needed

jraoult:
Sure @patrickkettner, I'll use the week upcoming week end for that and yes I'll make sure to not copy these big base64 guys over ;-)

patrickkettner:
sounds awesome!

thanks so much dude!

On Wed, Jan 28, 2015 at 10:33 PM, Jonathan Raoult <notifications@github.com>
wrote:

> Sure @patrickkettner <https://github.com/patrickkettner>, I'll use the
> week upcoming week end for that and yes I'll make sure to not copy these
> big base64 guys over ;-)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Modernizr/Modernizr/issues/1505#issuecomment-71963492>
> .
>



-- 
patrick

jraoult:
Sorry @patrickkettner it took me a little bit longer than expected to find the time work on it. Please see #1526

PR #1526, opened 2015-02-17
Add a test for javascript initiated video auto play

jraoult:
#### What's this PR do?

Adds a test that checks if a video can be started by javascript without user intervention (the JS call is not the consequence of a user initiated event)

#### Related issue
#1505


jraoult:
@patrickkettner This is definitely not a PR ready for merging but I wanted to check with you the validity of this approach.

Since `video/autoplay` and `video/javascriptautoplay` differ only by the way they start the video I extracted to code to a parametrizable builder that returns the test itself.

Another thing, I forked from `master` (214f02af7b6e030de5d6422133c36975f74929bb) but I couldn't manage to get `grunt build` working so I couldn't run the tests. Where can I find a working version of the build script?

patrickkettner:
@jraoult awesome work!

grunt build should be working - in what way is it not?

jraoult:
@patrickkettner if think this doesn't work since in the dependencies I have `build` is synchronous and returns a value instead of accepting a callback:
```
modernizr.build(config, function(output) {
  grunt.file.write(dest, output);
});
```

The result is that there is no `./dist/modernizr-build.js` file generated

patrickkettner:
actually, total screw up on my part. updating off master will give you a working version

jraoult:
@patrickkettner thanks, it works now :-)

jraoult:
@patrickkettner the results of my tests are disappointing because it seems that this test was not really required after all.  I can definitely confirm that autoplay (both by attribute and JS) passes for FF on Android and none of them pass for Chrome. So it seems that as of today autoplay triggered either by attribute or JS is treated the same way by browsers. I think I've been deceived by false negative for `videoautoplay` when I was prototyping this test earlier.

Bottom line we are testing two different use cases but the results are the same, does that make them redundant?

patrickkettner:
yeah, if there is no browser that does one but not the other, then the
detect doesn't really detect anything.

That being said, I would urge you to triple check whether or not your
original findings were correct. Id hate to loose a perfectly good detect

thanks again, regardless!

jraoult:
@patrickkettner no worries :-) yes, I'll check more thoroughly and I'll keep you posted

patrickkettner:
closing in the mean time, please do ping if your results turn out otherwise.

cheers!

jraoult:
Hey @patrickkettner, I found this issue in the FF bug tracker: https://bugzilla.mozilla.org/show_bug.cgi?id=659285

So it appears that it is possible in FF to disable "autoplay" as an attribute but still allow `play()` to be called in JS. This scenario would make my PR relevant after all :-)

PR #1533, opened 2015-02-23
Add docs badge to README

rrrene:
Hi there,

I want to propose to add this badge to the README to show off inline-documentation: [Inline docs](http://inch-ci.org/github/Modernizr/Modernizr)

The badge links to [Inch CI](http://inch-ci.org) and shows an evaluation by [InchJS](http://trivelop.de/inchjs), a project that tries to raise the visibility of inline-docs. Besides testing and other coverage, documenting your code is often neglected although it is a very engaging part of Open Source that encourages aspiring developers to jump into the source and see how it all ties together.

So far over 600 **Ruby** projects are sporting these badges to raise awareness for the importance of inline-docs and to show potential contributors that they can expect a certain level of code documentation when they dive into your project's code and motivate them to eventually document their own. I would really like to do the same for the **JavaScript** community and roll out support for JS over the coming weeks (early adopters are [forever](https://github.com/foreverjs/forever), [node-sass](https://github.com/sass/node-sass) and [when](https://github.com/cujojs/when)).

Although this is "only" a passion project, I really would like to hear your thoughts, critique and suggestions. Your status page is http://inch-ci.org/github/Modernizr/Modernizr

What do you think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,14 @@
   "bugs": {
     "url": "https://github.com/Modernizr/Modernizr/issues"
   },
+  "inch": {
+    "files": {
+        "included": [
+          "feature-detects/",
+          "src/"
+        ]
+      }
+  },
   "keywords": [
     "html5",
     "css3",
--- a/readme.md
+++ b/readme.md
@@ -1,4 +1,4 @@
-# Modernizr [![Build Status](http://img.shields.io/travis/Modernizr/Modernizr/master.svg)](http://travis-ci.org/Modernizr/Modernizr)
+# Modernizr [![Build Status](http://img.shields.io/travis/Modernizr/Modernizr/master.svg)](http://travis-ci.org/Modernizr/Modernizr) [![Inline docs](http://inch-ci.org/github/Modernizr/Modernizr.svg?branch=master)](http://inch-ci.org/github/Modernizr/Modernizr)
 
 ##### Modernizr is a JavaScript library that detects HTML5 and CSS3 features in the user’s browser.
 


ryanseddon:
That's neat but why do you have those other changes in your pr?

rrrene:
I do not have the slightest idea.. I used GitHub's web-interface to make the edit, nothing else.

Anyway, I rebased against the current master and updated the PR accordingly. Thanks for the tip!

ryanseddon:
@patrickkettner thoughts? I think this is great.

ryanseddon:
@rrrene the status page doesn't seem to parse any of the files in the feature-detects folder?

rrrene:
That could be included via a config file. I just did not put it in the initial commit because I did not want to force yet another config file into the root of your project.

But individual [files/directories can included and excluded](https://inch-ci.org/help/config_file_json) from the analysis.

patrickkettner:
@rrrene any chance inch could use package.json or bower.json fields rather than having its own?

patrickkettner:
for the record - I agree this is awesome, and am all for it

rrrene:
Now, I am not 100% sure I understand your suggestion: Is there an "include/exclude"-flag for package.json that I am not aware of or do you mean that we could use the package.json file to store the same config that would otherwise go to inch.json?

InchJS actually [already does recognize package.json and bower.json](https://github.com/rrrene/inchjs/blob/master/lib/docs/retriever.js#L37), but only if neither `lib/` nor `src/` are present ... leveraging this mechanism to load the inch config from a key inside package.json instead of its own file would indeed be trivial.

ryanseddon:
yeah would much prefer configuration to be in `package.json` under some key, much like browserify uses the `browser` key to store config.

patrickkettner:
@rrrene I meant what @ryanseddon elaborated on. A field in package.json that would contain everything normally in inch.json

rrrene:
Very good. I will implement that feature in the next version of InchJS and update the PR when it is done. :+1:

ryanseddon:
Thanks @rrrene look forward to landing this, great addition to the project.

rrrene:
Hi there, I completely forgot to *ping* this thread. :disappointed:  I added the discussed feature a week ago and updated this PR accordingly!

ryanseddon:
wow nice, missed your update.

PR #1535, opened 2015-02-24
improve npmignore

patrickkettner:
I goofed and published assets to the npm alpha branch when I didn't mean to
because I forgot npmignore didn't follow gitignore. adding more of the files to
here in order to prevent that from happening again.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,8 @@
 test
 media
+dist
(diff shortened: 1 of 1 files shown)


patrickkettner:
once this is merged, we should update bump to alpha2 in package.json and rerelease

PR #1536, opened 2015-02-25
fixes for appveyor

patrickkettner:
couple of things were going on. the `mocha.end` event was killing phantom
before the `__coverage__` object was fully uploaded sometimes, and so the JSON
would be invalid. secondly, the streams on windows 0.10 seemed to not always
call a `end`/`close` event for streams. I updated it to use the old fashioned
ta = data + chunk` method, which gets around that.

it built several times successfully, now hopefully this PR will as well...


ryanseddon:
boom, nice

PR #1537, opened 2015-02-26
fix vmin and vmax tests for windows browsers

patrickkettner:
firefox was failing on the test page because of the scrollbars on the page.

these detects are virtually identical - should we combine them?

### DIFF
--- a/feature-detects/css/vmaxunit.js
+++ b/feature-detects/css/vmaxunit.js
@@ -14,7 +14,7 @@
   }]
 }
 !*/
-define(['Modernizr', 'docElement', 'testStyles'], function( Modernizr, docElement, testStyles ) {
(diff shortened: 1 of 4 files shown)


patrickkettner:
in certain situations, browsers were off by one one this. I have seen that happen in a few places before, so I created the `roundedEquals` to check is if a number is equal to another, plus or minus 1.

I don't love the name - anyone got a better one?

patrickkettner:
@Modernizr/team-modernizr could someone give a once over to the changes? This is one of the last things blocking auto browser testing on a PR level

ryanseddon:
Can't think of a better name `approximateEquals`, that's probably worse. 

I'm happy to combine the two tests, can inject 4 elements and just extend the stylesheet that's injected.

PR #1545, opened 2015-02-28
fixes for ie 12

patrickkettner:
IE 12 failed because of its support for `webkit` selectors. considering more and more browsers are doing so, I moved `webkit` to the end of the line for sniffing. I can't imagine that it matters, but could @paulirish @SlexAxton @ryanseddon or @stucox confirm as to whether or not that order matters?


ryanseddon:
LGTM :+1:

PR #1551, opened 2015-03-09
Detect emoji support with various device pixel ratio

teebot:
Sampling a pixel to detect if it has a fill color could vary with screen
pixel density and thus return inaccurate results.

### DIFF
--- a/feature-detects/emoji.js
+++ b/feature-detects/emoji.js
@@ -10,12 +10,14 @@ Detects support for emoji character sets.
 define(['Modernizr', 'createElement', 'test/canvastext'], function( Modernizr, createElement ) {
   Modernizr.addTest('emoji', function() {
     if (!Modernizr.canvastext) return false;
-    var node = createElement('canvas'),
-    ctx = node.getContext('2d');
+    var pixelRatio = window.devicePixelRatio || 1;
+    var offset = 8 * pixelRatio;
+    var node = createElement('canvas');
+    var ctx = node.getContext('2d');
     ctx.fillStyle = '#f00';
     ctx.textBaseline = 'top';
     ctx.font = '32px Arial';
     ctx.fillText('\ud83d\udc28', 0, 0); // U+1F428 KOALA
-    return ctx.getImageData(16, 16, 1, 1).data[0] !== 0;
+    return ctx.getImageData(offset, offset, 1, 1).data[0] !== 0;
   });
 });


feature-detects/emoji.js
```diff
@@ -9,13 +9,16 @@ Detects support for emoji character sets.
 */
 define(['Modernizr', 'createElement', 'test/canvastext'], function( Modernizr, createElement ) {
   Modernizr.addTest('emoji', function() {
-    if (!Modernizr.canvastext) return false;
+    if (!Modernizr.canvastext || !window.devicePixelRatio) return false;
```
patrickkettner: we don't want to do that - while I can't think of a browser combination that would fail the second and not the first, it would still be an unnecessary heuristic. Instead, do something like

```

**Comment by patrickkettner** [comment]
awesome catch! thanks a ton for reporting it!

one small tweak, then we should be good to go!

**Comment by patrickkettner** [comment]
looks beautiful - only last thing is the squash the commits down to a single one, to keep the history clean.

then we are good to merge!

File: feature-detects/emoji.js [hunk 2]
```diff
@@ -10,12 +10,14 @@ Detects support for emoji character sets.
 define(['Modernizr', 'createElement', 'test/canvastext'], function( Modernizr, createElement ) {
   Modernizr.addTest('emoji', function() {
     if (!Modernizr.canvastext) return false;
+    var pixelRatio = window.devicePixelRatio || 1;
+    var offset = 8 * pixelRatio;
     var node = createElement('canvas'),
     ctx = node.getContext('2d');
```
ryanseddon: You're leaking a global here.

feature-detects/emoji.js
```diff
@@ -10,12 +10,14 @@ Detects support for emoji character sets.
...
     ctx.fillStyle = '#f00';
     ctx.textBaseline = 'top';
     ctx.font = '32px Arial';
     ctx.fillText('\ud83d\udc28', 0, 0); // U+1F428 KOALA
-    return ctx.getImageData(16, 16, 1, 1).data[0] !== 0;
+    return ctx.getImageData(offset, offset, 1, 1).data[0] !== 0;
   });
-});
+});
```
ryanseddon: Can you make sure you have a newline at eof.

feature-detects/emoji.js
dmethvin: I think it's just a style thing? `ctx` is okay because there's a comma on the previous line, it should have been indented. You might want to add JSCS to the build process for these issues now that `onevar` is removed from JSHint.

ryanseddon: oh totally missed the comma, @teebot can you just make that a `var`

ryanseddon: Also good idea @dmethvin about jscs I'll put something together.

teebot: Sure I'll make it a var. You're right that coma was easy to miss without
indentation.
I guess I'm on a different timezone, tomorrow I'll do that + blank line EOF
and squash down the commits to one.
Thanks!

patrickkettner:
LGTM!

patrickkettner:
thanks so much, dude!

feature-detects/emoji.js
hzoo: @ryanseddon I can help out with adding jscs - in master (and 0.12 which will be released soon) there's a new autofixing feature for whitespace which will help a lot with getting the codebase up to speed with the preferred style.

ryanseddon: @hzoo that would be great!

Issue #1552, opened 2015-03-10
SVG inline test false positive


eltonmesquita:
Recently i've gone through some false positive with the inline SVG test and found about using SVGRect for testing SVG support. So I've tested Modernizr and SVGRect to see which one is more reliable.
I've even written about it [here](http://codepen.io/eltonmesquita/blog/loading-inline-svg/) (I'm going to update it soon).

The problem is both methods give false positives in different browsers. I've tested both tests in some browsers and you can see the results and the tests [here](http://codepen.io/eltonmesquita/pen/GgXbvo/).

Although in my data both methods returned false positive in just one browser each, another Samsung device returned the wrong result in previous tests with Modernizr, but I decided to not include it because I don't have access to it anymore and can not retest. It seems like it is a problem related to some [Chromia](http://www.quirksmode.org/blog/archives/2015/02/counting_chromi.html) browsers that are shipped with some vendor's Android devices. The AOSP browser (the official Android browser) doesn't seem to have this issue.

This situation can be a problem, depending on the user target.

I was thinking about using both methods to try to detect the inline SVG support in a more relaible way, but I couldn't find a way.

Any thoughts on that?

patrickkettner:
So you are suggesting we return true if both methods work properly?

eltonmesquita:
If we do that, we would return less accurate results, but it would be at least false negatives and that is "less bad". It would be a trade-off, but more important this situation should be clear to the users, no matter what is the solution.

What do you think? Would it be a good trade-off?

patrickkettner:
Wanna PR the change?

PR #1553, opened 2015-03-11
Improve inline SVG detection

eltonmesquita:
Fix issue #1552


patrickkettner:
@eltonmesquita - I thought you said you did _not_ want to run both checks?

eltonmesquita:
Sorry, I've got a bit confused by your answer. But still, I think it is a better option, as it will not return any false positive, like the current test. If it is to be documented, where exactly this info should be? I don't see anything related to this kind of issue in the docs.

patrickkettner:
we usually add a `knownBugs` field to the detect if there are known problems.

feature-detects/svg/inline.js
```diff
@@ -7,8 +7,12 @@
   "notes": [{
     "name": "Test page",
     "href": "http://paulirish.com/demo/inline-svg"
+  }, {
+    "name: "Test page and results",
```
patrickkettner: you have a typo here that is breaking the parsing and thus the tests

`"name:` should be `"name":`

eltonmesquita: Sorry for that, ended up missing it. Fixed!

patrickkettner:
Could you squish down everything to a single commit? then we can get this landed :D

eltonmesquita:
Did I do it right, now? I don't know why the AppVeyour build failed, but it doesn't seem a problem with the code itself...

patrickkettner:
appveyor is shakey - it messed up during the npm install. I kicked off the build again, just to double check.

patrickkettner:
you would want to run `git rebase -i HEAD~7` and change all but the first `pick` to `f`s, then save and force push up the changes. That way the git history on Modernizr stays nice and clean

eltonmesquita:
Did I get it now? Got take a course on git...

patrickkettner:
@eltonmesquita seems like it got messed up :[

sorry again for complicated things. I cherrypicked and merged in via 60f6853567b4c2e47dd36007ee6a1a5de9304af5

thanks a ton for the PR!

eltonmesquita:
No problem, this mess happened because I lacked some Git skills and time to review everything, it wasn't your fault. Next time I contribute, I'll do it in a clean way!
Thanks a lot for helping and discussing!

patrickkettner:
:+1:

PR #1554, opened 2015-03-12
Fixed a regression on webkit family browsers with devicePixelRatio 1

teebot:
Shifting the image detection by a few pixels seems to solve a false negative on chrome and safari with devicePixelRatio 1.

### DIFF
--- a/feature-detects/emoji.js
+++ b/feature-detects/emoji.js
@@ -11,7 +11,7 @@ define(['Modernizr', 'createElement', 'test/canvastext'], function( Modernizr, c
   Modernizr.addTest('emoji', function() {
     if (!Modernizr.canvastext) return false;
     var pixelRatio = window.devicePixelRatio || 1;
-    var offset = 8 * pixelRatio;
+    var offset = 12 * pixelRatio;
     var node = createElement('canvas');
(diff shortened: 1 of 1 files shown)


ryanseddon:
An this still returns results correctly in all others?

patrickkettner:
what do you mean when you say fail? chrome for windows does not support
emoji, so a "fail" would mean you get a positive result there.

On Fri, Mar 13, 2015 at 3:25 AM, Thibaut Nguyen <notifications@github.com>
wrote:

> I just retested deviceRatio 1 and 2, it still fails for one browser:
> *Windows 8*
> IE 9 to 11 *pass*
> Firefox 36 *pass*
> Chrome 41 *fail*
>
> *OSX*
> Safari *pass*
> Firefox 36 *pass*
> Chrome 41 *pass*
>
> Is there a more reliable way to detect emojis than the print/sample method?
> The implementation of emojis is so different from an OS/Browser
> combination to the other that it's maybe worth splitting the heuristics by
> case. What do you think?
>
> —
> <https://github.com/Modernizr/Modernizr/pull/1554#issuecomment-78846773>.
>



-- 
patrick

PR #1562, opened 2015-03-22
handful of updates for new site

patrickkettner:


patrickkettner:
@ryanseddon could you give it a once over?

lib/build.js
```diff
@@ -5,9 +5,22 @@
 var inBrowser = typeof define == 'function' && typeof define.amd == 'object';
 /* jshint -W117 */
 
+var _extend = function(a, b) {
```
ryanseddon: Could we not just use lodash.extend? We're using it already in a few places.

patrickkettner: its not loaded at this point in the browser context

ryanseddon: Ah I see fair enough.

ryanseddon:
LGTM :+1:

PR #1564, opened 2015-03-25
CORS for IE9 & IE8

hypotenuse:
IE9 & IE8 formally have CORS support but it is implemented via `XDomainRequest` constructor. `XDomainRequest` is a simplified version of `XMLHttpRequest`.

### DIFF
--- a/feature-detects/cors.js
+++ b/feature-detects/cors.js
@@ -12,8 +12,9 @@
 }
 !*/
 /* DOC
-Detects support for Cross-Origin Resource Sharing: method of performing XMLHttpRequests across domains.
+Detects support for Cross-Origin Resource Sharing: the ability to make cross-domain requests.
 */
 define(['Modernizr'], function( Modernizr ) {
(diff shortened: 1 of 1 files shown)


patrickkettner:
Hiya @hypotenuse!

Thanks a bunch for opening up a PR, but I would be :-1: on this - this isn't CORS, its CORS-like.

For someone who doesn't look at the docs, and just looks at `Modernizr.cors`, this would be really misleading. 

thoughts @ryanseddon?

SlexAxton:
Yea, as a common CORS user, I'd agree, these are different enough to be dangerous. It *might* be worth pulling in an `XDomainRequest` test, and you could use both in your code explicitly.

patrickkettner:
+1 for separate XDomainRequest detect.

If you agree, @hypotenuse please open one up.

cheers!

SlexAxton:
(One quick concern would be that `XDomainRequest` seems very generic, even though it's actually very specific. So I'd love to make sure that it's well documented as an Old IE feature, not the generic 'cross domain request' functionality a browser might or might not have)

PR #1565, opened 2015-03-26
Add jscs to check style

hzoo:
@ryanseddon 
from https://github.com/Modernizr/Modernizr/pull/1551#discussion_r26175879.

I just copied the format from jshint so it will run it in on the same files (and same ignores).

I didn't know what the rules should be (so we can start discussing that now based on the code) so I just picked the google preset as a start (using the jscs autoconfigure option (`jscs --autoconfigure src/ lib/ feature-detects/`)

One option is to use it as a base or just copy the rules from the preset and add/remove as needed. If there isn't a written style guide or another one to follow we can just look at what's the majority.

Another thing would be how to go about fixing the styles after determining rules.
We can use the autofix setting `jscs --auto-fix` (only in master branch) to fix most of the whitespace changes and everything else would be manual. I went ahead and did the autofix changes in a seperate commit so the potential changes of using the google preset can be seen. I can remove the 2nd commit or make a new PR to actually fix the changes later.

And then adding it to the `grunt test` task after all that's fixed? Maybe that should be a separate issue? Right now you can do `grunt jscs` though after an npm install.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,0 +1,9 @@
+{
+	"preset": "google",
+	"disallowKeywordsOnNewLine": null,
+	"disallowMultipleVarDecl": null,
+	"disallowMultipleLineBreaks": null,
+	"maximumLineLength": null,
+	"requireCamelCaseOrUpperCaseIdentifiers": null,
+	"validateIndentation": null
+}
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -2,7 +2,7 @@
 
 var browsers = require('./test/browser/sauce-browsers.json');
 
-module.exports = function( grunt ) {
+module.exports = function(grunt) {
   'use strict';
 
   // Load grunt dependencies
@@ -48,6 +48,16 @@ module.exports = function( grunt ) {
           }
         ]
       }
+    },
+    jscs: {
+      src: [
+        'Gruntfile.js',
+        'src/*.js',
+        'lib/*.js',
+        'feature-detects/**/*.js',
+        '!src/html5printshiv.js',
+        '!src/html5shiv.js'
+      ]
     },
     jshint: {
       options: {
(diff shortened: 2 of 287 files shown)


patrickkettner:
@hzoo thank you so much for this!
Would you be able to rebase against master? A single commit for all of these changes would be best.

thanks again!

feature-detects/css/remunit.js
```diff
@@ -25,7 +25,7 @@ define(['Modernizr', 'createElement'], function( Modernizr, createElement ) {
     try {
       div.style.fontSize = '3rem';
     }
-    catch( er ) {}
+    catch ( er ) {}
```
ryanseddon: Can we add a rule here to make it drop the whitespace inside the parenthesis too?

hzoo: Sure - I'm just using the `google` preset as a base/start so yeah feel free for more suggestions in rules to add/remove.

patrickkettner: believe that would be `"disallowSpacesInsideParentheses": true`

patrickkettner: looks like google aligns pretty closely with our own style. it generates the least amount of style changes from all of the presets (crockford is 7k haha).

hzoo: Yeah I think I picked the one with the least errors (forgot)

ryanseddon:
Another thing we would want to add is a `lint` task that runs `jshint` & `jscs` much like the task in [jQuery](https://github.com/jquery/jquery/blob/master/Gruntfile.js#L159)

feature-detects/css/remunit.js
ryanseddon: Yep let's set that to true as we're pretty inconsistent on that front looking at the files.

patrickkettner:
@ryanseddon you mean something like `grunt.registerTask('lint', ['jshint', 'jscs'])`?

patrickkettner:
hah - didn't notice the link. agreed

ryanseddon:
Yes exactly like that and the our `test` just calls `lint`

feature-detects/css/remunit.js
hzoo: Interesting.. I don't know why but I had nulled the `disallowSpacesInsideParentheses` rule in the PR haha.

patrickkettner: while we are at it, `requireCurlyBraces` should be `true` as well

hzoo:
`grunt.registerTask('lint', ['jshint', 'jscs'])`

Ok so in the current `Gruntfile` https://github.com/Modernizr/Modernizr/blob/9e249d5a6b87ada3c3f7f5e5c419b9fc381d22aa/Gruntfile.js#L221

There's already the test and default tasks that run jshint. Should I change the references to run `lint` instead?

patrickkettner:
we would want to change that from `jshint` to `lint`, after creating the above mentioned `lint` task

Gruntfile.js
```diff
@@ -49,6 +49,16 @@ module.exports = function( grunt ) {
         ]
       }
     },
+    jscs: {
+      src: [
+        'Gruntfile.js',
```
hzoo: These are the same a jshint - any changes needed?

patrickkettner: should be good

feature-detects/css/remunit.js
hzoo: Ok so with that `if (elem === false) return props[i];`

would be

```js
if (elem === false) {
  return props[i];
}
```

patrickkettner: correctamundo

hzoo:
Autofix is great! Changed the requireCurly's manually since autofix only does whitespace currently (so those shouldn't break anything). Ran `gulp lint` with no issues.

patrickkettner:
since this commit modified the Grunt file, saucelabs won't run. I am running them locally just to confirm (I can't imagine it would break anything, but weirder crap has come up previously). Once it goes green I am good to merge :+1:

.jscsrc
```diff
@@ -0,0 +1,9 @@
+{
+	"preset": "google",
```
ryanseddon: Can you make this 2 spaces to match other rc files

patrickkettner:
sall good.

great work, @hzoo!

patrickkettner:
ugh, sorry @ryanseddon. I'll update real quick

ryanseddon:
no problem it's not a huge one

hzoo:
Cool! - I guess `editorconfig` didn't catch that since I must of copied another file with 4 spaces. There's also a `validateIndentation` rule although it seems to have some bugs - probably the most difficult rule jscs has.

PR #1570, opened 2015-04-01
cssvwunit, cssvhunit: test #modernizr width as a float and support round...

### LINKED ISSUE #1569: cssvwunit/cssvhunit tests give false negative in some browser zoom/widths
To reproduce (all latest browsers except Safari, more on that below):
1. Set your browser to 961px width
2. Zoom to 110%
3. http://jsfiddle.net/6L7n2243/show/

`window.innerWidth` and `window.innerHeight` appear to be _rounded_ by the browser based on the calculated viewport size after a zoom, rather than _floored_ like we're used to.

If the window is 961px wide and the zoom is 110%, the viewport width is 961 \* 1/1.1 = 873.63, and the innerWidth is 874. The test div with 50vw would then be 873.63/2 = 436.815. Which when floored is one less than innerWidth/2 because `floor(round(n)/2) == floor(n/2)` is not true for every n. There are several other zoom level/window width combinations to make this happen.

Problem is, Safari DOES floor innerWidth and innerHeight (which I think is right). That makes it impossible to just do `round(testDivWidthFloat * 2) == innerWidth` which would be really nice. The other solution I'm exploring is using `getComputedStyle(document.documentElement).width` which gives you the fractional value of the viewport, rather than `innerWidth`, but I can't find enough information to know if that is reliable or not.

chearon:
...ed, floored innerWidth and innerHeight

This fixes #1569 by assuming window.innerHeight, window.innerWidth may be rounded normally, floored, ceiled, etc by the browser. This test still fails in Safari when you're zoomed in, but it did before, and I have not been able to find away around it since Safari's vw/vh units stay the same as you zoom in.

### DIFF
--- a/feature-detects/css/vhunit.js
+++ b/feature-detects/css/vhunit.js
@@ -16,10 +16,10 @@
 !*/
 define(['Modernizr', 'testStyles', 'roundedEquals'], function( Modernizr, testStyles, roundedEquals ) {
   testStyles('#modernizr { height: 50vh; }', function( elem ) {
-    var height = parseFloat((window.getComputedStyle ?
+    var height = parseInt((window.getComputedStyle ?
                               getComputedStyle(elem, null) :
                               elem.currentStyle)['height'],10);
 
-    Modernizr.addTest('cssvhunit', roundedEquals(parseInt(height * 2, 10), window.innerHeight));
+    Modernizr.addTest('cssvhunit', roundedEquals(height, parseInt(window.innerHeight / 2, 10)));
   });
 });
--- a/feature-detects/css/vwunit.js
+++ b/feature-detects/css/vwunit.js
@@ -16,10 +16,10 @@
 !*/
 define(['Modernizr', 'testStyles', 'roundedEquals'], function( Modernizr, testStyles, roundedEquals ) {
   testStyles('#modernizr { width: 50vw; }', function( elem ) {
-    var width = parseFloat((window.getComputedStyle ?
+    var width = parseInt((window.getComputedStyle ?
                               getComputedStyle(elem, null) :
                               elem.currentStyle).width, 10);
 
-    Modernizr.addTest('cssvwunit', roundedEquals(parseInt(width * 2, 10), window.innerWidth));
+    Modernizr.addTest('cssvwunit', roundedEquals(width, parseInt(window.innerWidth / 2, 10)));
   });
 });


patrickkettner:
thanks a ton for the PR!

Rather than using Math directly, could you use roundedEquals like in
https://github.com/Modernizr/Modernizr/blob/16f5464df4220df7716c2fcb87c0a10e910e289d/feature-detects/css/vmaxunit.js
? That way we get both sides of the rouded coin

On Wed, Apr 1, 2015 at 3:58 PM, Caleb Hearon <notifications@github.com>
wrote:

> ...ed, floored innerWidth and innerHeight
>
> This fixes #1569 <https://github.com/Modernizr/Modernizr/issues/1569> by
> assuming window.innerHeight, window.innerWidth may be rounded normally,
> floored, ceiled, etc by the browser. This test still fails in Safari when
> you're zoomed in, but it did before, and I have not been able to find away
> around it since Safari's vw/vh units stay the same as you zoom in.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/Modernizr/Modernizr/pull/1570
> Commit Summary
>
>    - cssvwunit, cssvhunit: test #modernizr width as a float and support
>    rounded, floored innerWidth and innerHeight
>
> File Changes
>
>    - *M* feature-detects/css/vhunit.js
>    <https://github.com/Modernizr/Modernizr/pull/1570/files#diff-0> (5)
>    - *M* feature-detects/css/vwunit.js
>    <https://github.com/Modernizr/Modernizr/pull/1570/files#diff-1> (5)
>
> Patch Links:
>
>    - https://github.com/Modernizr/Modernizr/pull/1570.patch
>    - https://github.com/Modernizr/Modernizr/pull/1570.diff
>
> —
> <https://github.com/Modernizr/Modernizr/pull/1570>.
>



-- 
patrick

chearon:
Ah, did not see that! Yes I'll add the commit now. Looks like I will have to pass the float through parseInt first

chearon:
Is this good?

Side note: IMO `roundedEquals` should take an unrounded number and an integer. It doesn't make sense for it to check for `a-1` since a number is never rounded "down" (e.g 3.5 could become 3 or 4 but not 2).

patrickkettner:
@chearon I think you misunderstood the point of roundedEquals. Its not meant to replace `Math.floor`. Its meant to be used in situations where off-by-one is ok. This can happen with pixel rounding issues (where we may expect it to be 10px, but in one browser it is 9.4, and another browser its 10.5). 

That being said, your changes are the right direction, but I find them super hard to understand at a glance. Part of it is you never need to use a float at all, you can just `parseInt` straight off, and pass that result into `roundedEquals`.

chearon:
I do need to use a float when parsing `50vw` - that's the issue with #1569. Otherwise there are rounding errors if your browser is zoomed in. Maybe this isn't the place to use `roundedEquals`

patrickkettner:
you would use rounded equals on the int that is parsed, rather than parsing
a float.

On Fri, Apr 3, 2015 at 8:47 AM, Caleb Hearon <notifications@github.com>
wrote:

> I do need to use a float when parsing 50vw - that's the issue with #1569
> <https://github.com/Modernizr/Modernizr/issues/1569>. Otherwise there are
> rounding errors if your browser is zoomed in. Maybe this isn't the place to
> use roundedEquals
>
> —
> <https://github.com/Modernizr/Modernizr/pull/1570#issuecomment-89278765>.
>



-- 
patrick

chearon:
I get it now. But that does not work! If the div is 50.9px wide, the viewport would be at 101.8px = 102px in Chrome. If I parseInt the div, I get 50. Multiply that times 2 and I'm testing 100 against 102px. Not enough precision.

I think it's worth considering changing roundedEquals to accept a float and an int from the browser.

patrickkettner:
`parseInt` both values, compare them in roudnedEquals

chearon:
Guess I could pass `parseInt(window.innerWidth/2)` and not multiply the test div by 2...  I'll try that

chearon:
there ya go, sorry for being thick

Morriz:
Has this been merged? Safari for iOS still throws false positives ;(
I hope this fix solves it....

rejas:
Resolved via #2264

PR #1572, opened 2015-04-06
Don't include dotfiles in npm package

SimenB:
No reason for dotfiles to be included when `npm install`ing.
Running `npm pack`, and extracting before and after, then running [this](http://stackoverflow.com/a/16788549/1850276), shows that these files gets excluded.

```
.editorconfig
.gitattributes
.jshintignore
.jshintrc
.jshintrc
.npmignore
.travis.yml
```

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -6,3 +6,4 @@ appveyor.yml
 Gruntfile.js
 metadata.json
 modernizr.min.js
+.*


patrickkettner:
hiya @SimenB

Great idea! Thank you

SimenB:
@patrickkettner  Wow, quick. No problem :smile: 

A quick question, any reason you blacklist files (.npmignore) instead of whitelisting (`files` in package.json)? I generally finds having to explicitly include files keeps the package slimmer.

Docs on files: https://docs.npmjs.com/files/package.json#files

patrickkettner:
> any reason...

generally sucking at .npmignore.

We did not intend to publish it until 3.0 ships, but someone else, erm, did us the favor. As a result, I just threw something together quickly.

Does the `files` field accept globs?

SimenB:
`files` accepts folders, I'm actually unsure if it does glob. Easy enough to check though. If you run `npm pack` in the root of the project, the tarball produced is what gets included when someone `npm install`s your package

patrickkettner:
I'll check it out. Thanks again, dude!

PR #1573, opened 2015-04-06
Clean up usage of jshintrc

SimenB:
There were only 2 rules different in `lib/.jshintrc` vs the root `.jshintrc`.
I also tried to clean up some globals in tests.

I see you've been really good with comments. Want me to expand on the things in this PR?

EDIT: Annoying diff from GH on comments in "json"-files...

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -88,22 +88,8 @@
   "smarttabs": true,
 
   "globals": {
-    "_": true,
-    "$": true,
-    "Blob": true,
     "DocumentTouch": true,
     "Modernizr": true,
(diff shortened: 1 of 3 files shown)


patrickkettner:
I am confused about the `globals: null` change, could you explain?

SimenB:
It's to avoid inheriting any globals.
https://github.com/jshint/jshint/blob/master/src/cli.js#L536
It overrides any previous (http://underscorejs.org/#extend), so an empty object would probably do it as well. `null` is more explicit, I think

patrickkettner:
Thanks again, @SimenB!

PR #1576, opened 2015-04-11
remove redundant setClasses option

patrickkettner:
it is rather silly.


SlexAxton:
Ur rather silly

patrickkettner:
image

PR #1580, opened 2015-04-17
Fix bug when using modernizr in Safari with plugin ClickToFlash

mari-na-bzzz:
'Embed' element is not always direct child of body. This causes error when using modernizr in Safari with enabled plugin ClickToFlash.
This fix works fine for me but maybe problem is deeper (I can't understand how embed becomes not a direct child of body)

### DIFF
--- a/feature-detects/flash.js
+++ b/feature-detects/flash.js
@@ -26,7 +26,7 @@ define(['Modernizr', 'createElement', 'docElement', 'addTest', 'getBody'], funct
         bool.blocked = (result === 'blocked');
       }
       addTest('flash', function() { return bool; });
-      if (embed && body.contains(embed)) {
+      if (embed && embed.parentNode === body) {
         body.removeChild(embed);
       }
     };


patrickkettner:
can you post a page where this is happening for you? we are appending to the body, so it _should_ be...

mari-na-bzzz:
maps.yandex.ru

patrickkettner:
she's working for me - browser?

mari-na-bzzz:
Osx 10.9.5, Safari 7.1.2, you should turn on plugin ClickToFlash
When bug appears, users can't drag or zoom map

mari-na-bzzz:
I've found out that ClickToFlash simply finds every domElement with flash content and wraps them in his own div-elements. That's why modernizr crashes

patrickkettner:
would you be able to post a jsbin/codepen/whatever to test this? I can get
a fix in really quick that way.

On Sat, Apr 18, 2015 at 5:31 AM, mari-na-bzzz <notifications@github.com>
wrote:

> I've found out that ClickToFlash simply finds every domElement with flash
> content and wraps them in his own div-elements. That's why modernizr crashes
>
> —
> <https://github.com/Modernizr/Modernizr/pull/1580#issuecomment-94148280>.
>



-- 
patrick

patrickkettner:
fixed in b00aa00

patrickkettner:
thanks a ton for reporting!

mari-na-bzzz:
> would you be able to post a jsbin/codepen/whatever to test this? I can get
a fix in really quick that way.

This fix works fine, thanks)

PR #1608, opened 2015-06-22
Backport #1503, fixes #1431 in 2.x

### LINKED ISSUE #1431: Modernizer w/ CommonJS (Browserify)

sandermarechal:
This allows using 2.x in module loaders like Browserify

### DIFF
--- a/modernizr.js
+++ b/modernizr.js
@@ -1403,4 +1403,4 @@ window.Modernizr = (function( window, document, undefined ) {
 
     return Modernizr;
 
-})(this, this.document);
+})(window, window.document);


reinos:
+1 this will fix all issues with Browserify

hung-phan:
:+1:

jpetitcolas:
:+1:

patrickkettner:
thank you, but closing since its been released in v3.0 now

PR #1622, opened 2015-07-28
Fix detecting unprefixed gradients

### LINKED ISSUE #1621: gradient detection requires old gradient syntax
As reported in https://bugzilla.mozilla.org/show_bug.cgi?id=1176496#c22 , Modernizr's gradient detection depends on the old gradient syntax.  Its attempt to detect unprefixed gradients fails because it's testing legacy syntax that's only supported in prefixed form.  This means that if browsers ever remove prefixed gradients (which is seeming increasingly unlikely due to things like this), Modernizr will report that gradients are not supported.

Gradient syntax is described in https://drafts.csswg.org/css-images/#linear-gradients , and requires "to left top" rather than just "left top" in https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/gradients.js#L29

TRowbotham:
Use the "to" keyword when detecting the unprefixed version of CSS gradients.  This should fix #1621.

### DIFF
--- a/feature-detects/css/gradients.js
+++ b/feature-detects/css/gradients.js
@@ -30,7 +30,7 @@ define(['Modernizr', 'prefixes', 'createElement'], function(Modernizr, prefixes,
     var angle;
 
     for (var i = 0, len = prefixes.length - 1; i < len; i++) {
-      angle = (i == 0 ? 'to ' : '') + 'left top';
+      angle = (i === 0 ? 'to ' : '') + 'left top';
       css += str1 + prefixes[i] + 'linear-gradient(' + angle + ', #9f9, white);';
     }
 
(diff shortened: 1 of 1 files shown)


TRowbotham:
appveyor seems to be broken right now :(

feature-detects/css/gradients.js
```diff
@@ -26,10 +26,14 @@ define(['Modernizr', 'prefixes', 'createElement'], function(Modernizr, prefixes,
 
     var str1 = 'background-image:';
     var str2 = 'gradient(linear,left top,right bottom,from(#9f9),to(white));';
-    var str3 = 'linear-gradient(left top,#9f9, white);';
+    var css = '';
+    var angle;
+
+    for (var i = 0, len = prefixes.length - 1; i < len; i++) {
+      angle = (i === 0 ? 'to ' : '') + 'left top';
```
ryanseddon: You could just move `'left top'` string to the line below e.g.

```js
angle = (i === 0 ? 'to ' : '');
css += str1 + prefixes[i] + 'linear-gradient(' + angle + 'left top, #9f9, white);';
```

PR #1667, opened 2015-09-21
add prefixedCSSValue, to test for prefixed css values

### LINKED ISSUE #688: Feature request: prefixed for linear gradients
As the linear-gradient background property has its own prefixes for every browser, it would be useful to have something like Modernizr.prefixed('linear-gradient')

patrickkettner:
fixes #688

I hate the name, but I can't think of something better. @ryanseddon?


ryanseddon:
lgtm can't think of a better name either, it's fairly descriptive of what it does so :+1:

PR #1669, opened 2015-09-21
Add msedge to sauce config.

jdalton:
Add msedge to sauce config using the short name for the platform (will see if that works).

### DIFF
--- a/test/browser/sauce-browsers.json
+++ b/test/browser/sauce-browsers.json
@@ -1,5 +1,12 @@
 [
   {
+    "browserName": "microsoftedge",
+    "platform": "WIN10",
+    "version": "20.10240"
+  },
+
+
+  {
     "browserName": "internet explorer",
     "platform": "XP",
     "version": "7"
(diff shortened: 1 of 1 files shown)


test/browser/sauce-browsers.json
```diff
@@ -1,5 +1,12 @@
 [
   {
+    "browserName": "microsoftedge",
+    "platform": "WIN10",
+    "version": "20.10240"
```
patrickkettner: can we leave off `version`? normally sauce will default to `latest` when this is done, no?"

jdalton: Maybe, though why would we be any different than Chrome/Firefox?

I've got  a thread going with them to change the version from `20.x` to `12.x` based on our [blog guidance](http://blogs.windows.com/msedgedev/2015/09/21/understanding-versions-in-an-evergreen-browser/).

patrickkettner: on our config?

it is behind, but it is supposed to represent the stable, beta, dev, and canary versions. if sauce only supports stable, then the default version would be ideal.

jdalton: I think experimenting with dropping version numbers would be more approp for another pr.

Issue #1597, opened 2015-05-19
Can't require('modernizr')


ScreenDriver:
I am using [webpack](http://webpack.github.io) to modularize and bundle my code (it's like [browserify](http://browserify.org) but with some advantages). I installed Modernizr `3.0.0-alpha4` via [npm](https://www.npmjs.com) and required Modernizr in my code. Unfortunately I get following error:

```sh
1 vagrant@vagrant-ubuntu-trusty-64 /vagrant % webpack                                                      :(
Hash: d313d14e909b8cd939c0
Version: webpack 1.9.7
Time: 894ms
    Asset    Size  Chunks             Chunk Names
bundle.js  717 kB       0  [emitted]  main
    + 173 hidden modules

ERROR in ./~/modernizr/package.json
Module parse failed: /vagrant/node_modules/modernizr/package.json Line 2: Unexpected token :
You may need an appropriate loader to handle this file type.
| {
|   "name": "modernizr",
|   "version": "3.0.0-alpha.4",
|   "description": "Modernizr is a JavaScript library that detects HTML5 and CSS3 features in the user’s browser.",
 @ ./~/modernizr/lib/build.js 131:12-38

ERROR in ./~/modernizr/lib/build.js
Module not found: Error: Cannot resolve module 'generate' in /vagrant/node_modules/modernizr/lib
 @ ./~/modernizr/lib/build.js 147:2-72

ERROR in ./~/modernizr/lib/build.js
Module not found: Error: Cannot resolve module 'lib/generate-banner' in /vagrant/node_modules/modernizr/lib
 @ ./~/modernizr/lib/build.js 147:2-72

ERROR in ./~/modernizr/lib/build.js
Module not found: Error: Cannot resolve module 'package' in /vagrant/node_modules/modernizr/lib
 @ ./~/modernizr/lib/build.js 147:2-72

ERROR in ./~/modernizr/lib/metadata.js
Module not found: Error: Cannot resolve module 'fs' in /vagrant/node_modules/modernizr/lib
 @ ./~/modernizr/lib/metadata.js 1:9-22

ERROR in ./~/modernizr/lib/polyfills.json
Module parse failed: /vagrant/node_modules/modernizr/lib/polyfills.json Line 2: Unexpected token :
You may need an appropriate loader to handle this file type.
| {
|   "html5shiv": {
|     "name": "HTML5 Shiv",
[...]

PR #1697, opened 2015-10-03
Include built modernizr in npm package

SimenB:
Also add browser field to allow requiring it.

This allows me to just do `import 'modernizr'` in a webpack application (probably browserify as well), and get the built file. This way I don't have to keep moderizr in my source code, or do any configuration

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,5 @@
 test
 media
-dist
 gh-pages
 appveyor.yml
 Gruntfile.js
--- a/package.json
+++ b/package.json
@@ -38,10 +38,12 @@
     "proxyquire": "1.7.2"
   },
   "scripts": {
+    "prepublish": "grunt build",
     "test": "grunt test --stack"
   },
   "main": "./lib/cli",
   "bin": "./bin/modernizr",
+  "browser": "dist/modernizr-build",
   "repository": {
     "type": "git",
     "url": "git://github.com/Modernizr/Modernizr.git"


patrickkettner:
thanks for the PR!
but there isn't a dist folder, nor is there a modernizr-build by default. This isn't meant to be required directly.

thanks, though!

SimenB:
`"prepublish": "grunt build"` creates the dist directory when `npm publish` is run

patrickkettner:
we don't want people using the dev file. it is awful performance for the web. There will be a better way to use modernizr with npm very soon, though

SimenB:
Ah, ok. Is there an issue I can track for that?

patrickkettner:
nah, ill be PRing it tomorrow, most likely. its the bower builder, just slightly modified to work with npm as well.

SimenB:
Ah, so _really_ soon. Looking forward to it!

patrickkettner:
thanks again for the PR, always very appreciated

SimenB:
Kinda related, any news on https://github.com/Modernizr/Modernizr/issues/1597#issuecomment-110421453? Will that be supported?

patrickkettner:
its something Id love to see happen, sooner rather than later. no public
plans in the immediate future, however. there a fair amount of maintenance
that feels a bit more important short term.

On Sat, Oct 3, 2015 at 3:06 AM, Simen Bekkhus <notifications@github.com>
wrote:

> Kinda related, any news on #1597 (comment)
> <https://github.com/Modernizr/Modernizr/issues/1597#issuecomment-110421453>?
> Will that be supported?
>
> —
> <https://github.com/Modernizr/Modernizr/pull/1697#issuecomment-145233647>.
>



-- 
patrick

SimenB:
Aight, thanks for the update!

alexpalombaro:
I couldn't find a way to build modernizr with webpack so ended up creating a (plugin)[https://github.com/alexpalombaro/modernizr-webpack-plugin] that did some of what I wanted. Not as nice as just being able to simply use `import 'modernizr/requiredtest'` but being able to custom build modernizr during webpack compile (and inject into html-webpack-plugin) has made things a little easier.

SimenB:
@alexpalombaro Ah, nice! I'll try to find the time to give it a whirl :smile: 

@patrickkettner any news on that npm usage? :grinning: And will it require a connection to the internet (if it's the same as the bower thing with added `.tar.gz`, I suppose so)?

patrickkettner:
@SimenB yes, it would require a internet connection. Is there something on npm that doesn't require an internet connection to install?

SimenB:
We are using a proxy of npm (Nexus) as our CI-server has no direct access to the net for legal reasons.

patrickkettner:
Ah. Been there. Then I would probably suggest creating a `postInstall` script that generates your build of Modernizr from the standard npm package

SimenB:
Yeah, we'll probably keep a premade file in our sources, or go for @alexpalombaro's plugin above :smile:

alexpalombaro:
@SimenB not sure about `htmlWebpackPluginsOptions` ... currently it works by monkey patching the method `htmlWebpackPluginAssets` and including the modernizr bundle in the resulting assets list. This way as far as html-webpack-plugin is concerned its just another linked asset so all the templating features of the plugin should function the same.

I'm not a huge fan of monkey patching but I wasn't able to figure out any other means.

jbrantly:
> nah, ill be PRing it tomorrow, most likely

I took a quick look but didn't see this.

synthecypher:
Can someone justify why we can't just update Modernizr to support `require` without having to build the library ourselves?
I understand the point of Modernizr is to configure the build to only output the tests you need, but why not just expose the tests as individual modules so they can be individually imported where testing is required.
Like @alexpalombaro said using `import 'modernizr/requiredtest`.

patrickkettner:
The "justification" is that that is what we are doing. The work just hasn't been done.

synthecypher:
@patrickkettner Sorry I didn't realize it was still on the roadmap.

jpgcode:
This is starting to be a blocker for using Modernizr with a module system. Please provide any clear workaround in documentation. Can´t make it work with webpack.

BillyFigueroa:
It has basically been a year and like 4 months since this was brought up and still very little information has been put out on a release date or the current status. Modernizr is going to take a big L if this is not added

The 2 biggest things in development right now are react and angular and both use require/ import syntax. You are blocking a bunch of developers from being able to use this library with out a nasty hack

patrickkettner:
There is an open PR with task items you are more than welcomed to lend a hand on

SimenB:
@patrickkettner this: #2108?

PR #1700, opened 2015-10-05
Add Bower File

carmona:
adding bower.json file helps everyone using modernize as a bower component (currently we have to manually edit the files in order to use it)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,25 @@
+{
+  "name": "modernizr",
+  "homepage": "https://github.com/Modernizr/Modernizr",
+  "version": "3.0.0",
+  "main":[
+    "src/Modernizr.js"
+  ],
+  "license": "MIT",
+  "ignore": [
+    "media",
+    "test",
+    "*.md",
+    "Gruntfile.js",
+    "package.json",
+    "*.yml"
+  ],
+  "keywords": [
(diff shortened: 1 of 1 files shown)


patrickkettner:
Hey @carmona!

Thanks a lot for the pull request - but we actually don't want to do this. There is not a single modernizr file as there was previously. We offer a bower downloading service rather than publishing to the official repo.

All you need to do is grab the build URL found at the top of any generated modernizr file 

screenshot of a modernizr build, higlighting the build url

and add a `.tar.gz` to the end of it.

Now you can add that to your `bower.json`, run the requisite install, and you have a built version ready for processing and deployment. 

`bower install --save https://modernizr.com/download?-cssanimations-csscalc-csscolumns-csstransforms.tar`

cheers, and thanks again

carmona:
Well, thanks for the response but this method is not actually working. Are you sure it should load a tarball?

captura de tela 2015-10-05 as 15 29 20

Both `curl` and `wget` also get the whole html file for the download page.

patrickkettner:
Remove the / after "download"

carmona:
Got it working now, thanks a lot! Actually got a smaller build with just the features needed.

patrickkettner:
awesome - thats why we do it :]

PR #1711, opened 2015-10-15
fix crlf (windows) end of lines forcing to lf (unix)

deformhead:


patrickkettner:
@deformhead thanks for the PR!
is this an issue somewhere, or a preventative measure

deformhead:
Both. We use it for a project at work, and we need to build a vendors file with all libraries installed with npm by a Jenkins (unix) server. Since some javascript files are in CRLF, the build fails.

So if someone works on this kine of environment, it may be a problem.

patrickkettner:
Sure, what I am trying to say is that this will only prevent it from happening in the future. If you see this issue on a file(s) in the repo currently - what are they, and can you update them within this PR?

deformhead:
Ok, my mistake : non-LF files was just media files ( .ai, .eps, .pdf and .png ), so no worries.

Besides, I updated my PR to prevent LF files to be changed ( preventative measure, so ).

ryanseddon:
Could we just force all to LF rather than target certain file extensions?

```bash
* text eol=lf
```

deformhead:
That was my first call, but media files was seen as text ( converting media files to LF brake them ).

So, two options :

I target files to be converted like I have done, but making a list is ugly, I agree.

Otherwise, I keep my first idea to keep LF on all text files, but I need to be certain which files do you want to "escape" from being a text ( which need a listing too, but smaller ). Only media files ? Most important, we have to keep in mind that we need to update this .gitattributes file if another media type come to avoid it to be convert and to be broken.

So, option one is ugly but safier and option two is better but tricky.

patrickkettner:
How are you getting the files on your build server

deformhead:
What do you mean ?

The fail on our build is not from a Modernizr mistake, at the end, but from another library, sorry for my first comment.

Now, I'm only dealing with forcing text files ( and text only ) to be LF, in this PR.

patrickkettner:
we should be able to use a `.gitattributes` file directly in the `media/` folder. it will only apply to those files. then keep your original change in the main `.gitattributes` file

deformhead:
@patrickkettner, I updated my PR in this way.

ryanseddon:
Thanks @deformhead

PR #1757, opened 2015-11-14
fix nthchild false negative

### LINKED ISSUE #1675: nth-child false negative
I'm seeing a `no-nthchild` class added in browsers that support nth-child.

No errors in the console, but adding a breakpoint in to [line 35](https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/nthchild.js#L35) of that test reports the `offsetWidth` of each `div` to be `0` (should be `1` even if not supported).

Using the following build of modernizr: https://modernizr.com/download?flexbox-flexboxlegacy-flexboxtweener-flexwrap-nthchild-dontmin-setclasses-shiv

Tested: Chrome/Firefox latest on Ubuntu 15.04, IE11/Edge on Windows 10

mstrutt:
Passing a function in to `Modernizr.addTest` resulted the code being executed *after* `elem` had been removed from the DOM (and so had no layout). I've moved the code to be executed immediately and the just the result  (which I renamed just to make the last line a bit more human readable) passed into `Modernizr.addTest`.

I searched the rest of the css feature-detects and nthchild appears to be the only test with this issue. That said, might be worth adding a little note to the docs about time of execution with `testStyles` and `addTest`.

All tests passed locally before *and* after this fix, but it was a bit beyond me to dive into what's happening with them

fixes #1675

### DIFF
--- a/feature-detects/css/nthchild.js
+++ b/feature-detects/css/nthchild.js
@@ -27,15 +27,12 @@ define(['Modernizr', 'testStyles'], function(Modernizr, testStyles) {
   // A Javascript loop then tests if the `<div>`s have the expected width
   // using the modulus operator.
   testStyles('#modernizr div {width:1px} #modernizr div:nth-child(2n) {width:2px;}', function(elem) {
-    Modernizr.addTest('nthchild', function() {
-      var elems = elem.getElementsByTagName('div'),
-      test = true;
+    var elems = elem.getElementsByTagName('div'),
+    correctWidths = true;
 
-      for (var i = 0; i < 5; i++) {
-        test = test && elems[i].offsetWidth === i % 2 + 1;
-      }
-
-      return test;
-    });
+    for (var i = 0; i < 5; i++) {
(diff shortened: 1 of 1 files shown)


feature-detects/css/nthchild.js
```diff
@@ -27,15 +27,12 @@ define(['Modernizr', 'testStyles'], function(Modernizr, testStyles) {
   // A Javascript loop then tests if the `<div>`s have the expected width
   // using the modulus operator.
   testStyles('#modernizr div {width:1px} #modernizr div:nth-child(2n) {width:2px;}', function(elem) {
-    Modernizr.addTest('nthchild', function() {
-      var elems = elem.getElementsByTagName('div'),
-      test = true;
+    var elems = elem.getElementsByTagName('div'),
+    correctWidths = true;
```
ryanseddon: Can you change this to multiple `var`'s this is an old one but all other tests should use this pattern going forward

mstrutt:
done :)

ryanseddon:
Thanks @mstrutt

PR #1760, opened 2015-11-16
Make Flash Test element invisible to prevent flickering

nicam:
Hi There!

We had a problem that the element that is attached to the body to detect flash was visible if the page has a background that is not white.

Cheers!

### DIFF
--- a/feature-detects/flash.js
+++ b/feature-detects/flash.js
@@ -66,6 +66,7 @@ define(['Modernizr', 'createElement', 'docElement', 'addTest', 'getBody', 'isSVG
       var inline_style;
 
       embed.type = 'application/x-shockwave-flash';
+      embed.setAttribute('style', 'display: none;');
 
       // Need to do this in the body (fake or otherwise) otherwise IE8 complains
       body.appendChild(embed);


niksy:
Maybe this is better achieved with `visibility: hidden;`, display property can sometimes cause unusual behavior.

nicam:
Yeah I wasn't sure whether or not it was necessary but I'm fine with both solutions

patrickkettner:
thanks for the pull request, @nicam!
We wouldn't want to use `display: none`, because that is what crappier flash blockers do to "hide" the embed. i'd prefer a `height: 1px; width: 1px;` sort of thing

nicam:
@patrickkettner hmm, what about `height: 1px; width: 1px; position: absolute; left: -9999px; top: -9999px;`

patrickkettner:
his the same issue of that being a standard way to hide stuff. Is the 1px for several milliseconds causing an issue?

nicam:
It will still cause a flicker although a tiny one. But as far as I see further below it's checking for a change from a flash blocker by testing if the cssText has changed:

`````
          inline_style = embed.style.cssText;
          if (inline_style !== '') {
            // the style of the element has changed automatically. This is a
            // really poor heuristic, but for lower end Flash blocks, it the
            // only change they can make.
            runTest('blocked', embed);
          }
```
As long as we use some weird number instead of -9999px then the CSS text will still be different from the value we have set and we can still identify it.

Something like:
`height: 1px; width: 1px; position: absolute; left: -7432px; top: -8311px;`

Plus we also have to change the check in the if to:
`if (inline_style !== 'height: 1px; width: 1px; position: absolute; left: -7432px; top: -8311px;') {`

This would result in no flickering but it would also trigger blocked when the css was changed from a cheap flash blocker.

WDYT?

PR #1772, opened 2015-11-24
Fix build query not including setclasses when passed as an option

akamike:
I stumbled upon this when moving from a manually built Modernizr file to a bower installed one. The build URL provided in the file did not include the `setclasses` property, so installing through bower as a tar.gz omitted this feature - despite the fact the original file used it.

This bug can be demonstrated on the current https://modernizr.com/download site itself:

1. Pick your feature detections
2. Ensure "Add CSS classes" is checked
3. Build and view the resulting file
4. Build URL in the header comment is missing `setclasses`

Example following these steps using `inlinesvg` generates this codepen: https://codepen.io/anon/pen/KdLxWb

This is happening as the generate script is directly modifying config to remove the setClasses special case option, which means it no longer exists for the rest of the build process (e.g. build-query) to work with.

Another part of this is even if the property was being set, it is using the wrong label anyway, so would still fail to include the `setClasses` option. This PR fixes that too.

Its usefulness doesn't just lie with copying the build URL for bower, it also makes it more consistently clear what features were intended to be included. Currently, the "bad" URLs work on the site as "Add CSS classes" is checked by default and replaced into the browser history state - shouldn't really depend on that behaviour.

### DIFF
--- a/lib/build-query.js
+++ b/lib/build-query.js
@@ -27,9 +27,6 @@ define(['lodash', 'metadata'], function(_, metadata) {
     // A few of the values have to be massaged in order to match
     // the `value`
     var opts = _.map(config.options, function(opt) {
-      if (opt == 'setClasses') {
-        return 'cssclasses';
-      }
       if (opt.indexOf('html5') === 0) {
         opt = opt.replace('html5', '');
       }
--- a/src/generate.js
+++ b/src/generate.js
@@ -23,9 +23,6 @@ define(['lodash'], function(_) {
     // Some special cases
     var setClasses = _.contains(config.options, 'setClasses');
 
-    // Remove the special cases
-    config.options = _.without(config.options, 'setClasses');
-
     var output = 'require(["ModernizrProto", "Modernizr", "testRunner"';
 
     // Needed named module requires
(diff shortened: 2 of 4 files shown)


ryanseddon:
Amazing, thanks!

ryanseddon:
@patrickkettner will this kick off a new deploy for the site or do I need to trigger it?

patrickkettner:
already set it off

On Thu, Nov 26, 2015 at 6:41 PM, Ryan Seddon <notifications@github.com>
wrote:

> @patrickkettner <https://github.com/patrickkettner> will this kick off a
> new deploy for the site or do I need to trigger it?
>
> —
> <https://github.com/Modernizr/Modernizr/pull/1772#issuecomment-160030869>.
>



-- 
patrick

akamike:
@patrickkettner I checked the site to see the change in action, however it is still exhibiting the original issue. Was this deployed? Running https://github.com/Modernizr/modernizr-neue locally works fine.

patrickkettner:
try now?

On Wed, Dec 2, 2015 at 6:23 AM, Mike Robinson <notifications@github.com>
wrote:

> @patrickkettner <https://github.com/patrickkettner> I checked the site to
> see the change in action, however it is still exhibiting the original
> issue. Was this deployed? Running
> https://github.com/Modernizr/modernizr-neue locally works fine.
>
> —
> <https://github.com/Modernizr/Modernizr/pull/1772#issuecomment-161312938>.
>



-- 
patrick

akamike:
All good now, thanks! :)

PR #1778, opened 2015-11-27
add detect for css snap points

### LINKED ISSUE #1777: Request: CSS Scroll Snap Support
Hi there,

Love your library, it's worked really well for me in the past.

Just a quick thought / question; Would you be able to add a feature test into Modernizer that tests for the support for CSS Scroll Snap?: http://caniuse.com/#feat=css-snappoints

It works perfectly in FireFox and iOS Safari (9) and average in IE10/11. 

I've managed to get detection to work by doing something like this (though I'm happy for feedback on how to make the test a bit more robust)

```
  //detect if we support scroll snap or not
  var scrollSnap = false;
  scrollSnap = 'webkitScrollSnapType' in document.documentElement.style;
  scrollSnap = 'scrollSnapType' in document.documentElement.style;
```

That test works but I'd love to keep all my detections inside Modernizer so I can keep things organised. 

Here is a CodePen I've been working on that displays a scroll snap interface

http://codepen.io/simoncodrington/pen/ojrwrN

Thanks again for your library.

patrickkettner:
fixes #1777


simonrcodrington:
Thanks for the quick reply @patrickkettner :)

Hopefully this helps someone who needs a quick way to detect snap scrolling.

Cheers.

PR #1786, opened 2015-12-02
Implement browserify support

tellnes:
Hi

Here is a working implementation for Browserify without changing anything. It lets you require just the tests you want and the Browserify builder will do it's magic.

Usage example:
```js
require('modernizr/feature-detects/svg/inline') // inlinesvg
require('modernizr/feature-detects/touchevents') // touchevents

const Modernizr = require('modernizr')
Modernizr.init({ classPrefix: 'modernizr-' })

console.log(Moderizr.touchevents)
```

I've created one new methods called `Modernizr.init()`. This method must called before the user starts checking the results of the tests. The reason for this method is that we need a way to configure Modernizr before any tests are run. In most situations, `process.nextTick(Modernizr.init)` will be the best option. 

I was considering a `cleanup` or `finish` method instead of the `init` method and just running `Modernizr._q` there, but that does not solve the configuration problem.

### DIFF
--- a/browserify/main.js
+++ b/browserify/main.js
@@ -1,0 +1,33 @@
+var ModernizrProto = require('../src/ModernizrProto');
+var Modernizr = require('../src/Modernizr');
+var testRunner = require('../src/testRunner');
+var setClasses = require('../src/setClasses');
+var classes = require('../src/classes');
+var addTest = require('../src/addTest');
+
+module.exports = Modernizr;
+
+var hasRun = false;
+ModernizrProto.init = function(options) {
+  if (hasRun) return;
+  hasRun = true;
+
+  options = options || {};
+
+  if (options.classPrefix);
+    Modernizr._config.classPrefix = options.classPrefix;
+
+  testRunner();
+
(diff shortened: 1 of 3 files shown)


patrickkettner:
Hi @tellnes!
I am actually planning on switching out the module structure from require based to es6 modules in the next couple of weeks. how much would that change this contribution?

tellnes:
Most of the `browserify/transform.js` code will probably need to be changed. But you probably still need to solve what I'm solving in `browserify/main.js` with `Modernizr._q` and configuration.

The reason I implement a custom transform instead of using eg. `deamdify` is because of what I'm doing in the `rewriteDependencyPath` function. Modernizr expects all paths to be relative to the `src/` directory, while I need them relative to their own filepath.

ryanseddon:
@patrickkettner I hope you'll save yourself some pain and use jscodeshift to automagically convert to es6.

patrickkettner:
To do the port, yeah. Hopefully it will be mostly just cleanup

PR #1794, opened 2015-12-04
Add test for es6 arrow functions

vincentriemer:
Following the conventions of the ES6 Generators test, added a new test for the ES6 arrow function syntax. 

I manually tested the code within the try-catch block and it succeeds in the latest Chrome, Firefox, and Edge but fails in the latest Safari which lines up with [this support table](https://kangax.github.io/compat-table/es6/).

### DIFF
--- a/feature-detects/es6/arrow.js
+++ b/feature-detects/es6/arrow.js
@@ -1,0 +1,22 @@
+/*!
+{
+  "name": "ES6 Arrow Functions",
+  "property": "arrow",
+  "authors": ["Vincent Riemer"],
+  "tags": ["es6"]
+}
+!*/
+/* DOC
+Check if browser implements ECMAScript 6 Arrow Functions per specification.
+*/
+define(['Modernizr'], function(Modernizr) {
+  Modernizr.addTest('arrow', function() {
+    try {
+      /* jshint evil: true */
+      eval('()=>{}');
+    } catch (e) {
+      return false;
+    }
+    return true;
+  });
+});


patrickkettner:
could you using `eval('()=>{}')`? shave a few bits

vincentriemer:
I'm fine with that, in fact that was what I originally planned to do but I came across the generator test code decided instead to follow an existing pattern instead. 

@mikach is there any reason why you went with a Function constructor instead of using `eval`?

patrickkettner:
@vincentriemer sorry this got lost in the holidays!

could you squish down to a single commit?

vincentriemer:
Done!

patrickkettner:
thanks!

kamilbrk:
Apologies for commenting in the old thread, just want to comment on what happened with our product today due to this change.

We use CI and 'gulp-moderizr', which depends on 'customizr', which in the end depends on 'Modernizr' through npm. Thing is, that 3rd level dependency is not locked to a specific version, so the 3.4.0 release from yesterday evening slipped into our product through CI.

This check for arrow functions is defined as 'arrow', so far so good.

It adds Modernizr.arrow and 'arrow' class to <html> element, as expected, right?

I know that I could have added a custom prefix for Modernizr CSS classes on <html>, but since I didn't, it did style my entire app (<html> element) as an arrow, because we had some .arrow {} selectors in CSS.

PSA: Please check your styles, please check your CI configs, consider Yarn instead of NPM to lock down deep dependencies, etc. We spent quite a while trying to figure out why our page is wrapped in a tiny box, even though QA was fine.

ryanseddon:
Thanks for the comments, that's a tough one to mitigate when transient
dependencies aren't version locked.

Issue #1427, opened 2014-09-22
False negative on Video Autoplay (Modernizr V3)


germanalvarez:
After checking how did video Autoplay work on v3, I just opened my website from Firefox and it had a no-videoautoplay class, this is a false negative as it does support it:

![screenshot_1](https://cloud.githubusercontent.com/assets/8863375/4359358/2dc8370c-4271-11e4-9579-dab1600c30aa.jpg)

On Chrome, there's another false positive (no-videoautoplay).
On IE11, there's another false positive (no-videoautoplay).
On Opera, it detects it correctly (videoautoplay).

ryanseddon:
I believe I've found the fix wanna [test this](https://codepen.io/anon/pen/QyWQjQ) and see if you get the correct results?

ryanseddon:
Closed via #1795

yuchant:
Chrome 48.0.2564.116 unreliably detected. First load is OK, subsequent loads shows no-videoautoplay class.

petrklus:
Same on Chrome 55

glomotion:
Chrome 84.0.4147.105 in Mac 10.15.6 is incorrectly reporting false:
https://codepen.io/glomotion/pen/gOrpoXG
<img width="795" alt="Screen Shot 2020-08-10 at 11 30 08 am" src="https://user-images.githubusercontent.com/1452237/89746432-eb267000-dafc-11ea-883f-dc4effc9b44c.png">

PR #1795, opened 2015-12-05
Add a new base64 encoded h264 video to fix autoplay false negative in Safari

### LINKED ISSUE #1692: false negative for videoautoplay in Safari 9.0
https://codepen.io/anon/pen/meRajz
Videos still autoplay.

ryanseddon:
Seems we had an encoding issue with our h264 test video, created a new video and this seems to fix the problem it's only slightly bigger than the last but has the benefit of actually working :trollface: 

Test case: https://codepen.io/anon/pen/QyWQjQ

Fixes #1692, #1427


patrickkettner:
:+1: - thanks ryan!

patrickkettner:
@ryanseddon good to merge?

MarvelousM:
Great!!! :) Made my day

PR #1807, opened 2015-12-14
fix(src/generate): avoid prepending 'test/' to absolute paths

### LINKED ISSUE #1799: No longer possible to pass path to custom tests to Modernizr.build
It's no longer possible to pass the paths to custom tests to Modernizr.build, as it fails trying to load the property while generating the buildQuery. This used to work with previous versions of Modernizr v3.

terinjokes:
If we detect that the user has provided an absolute path, avoid
prepending 'test/' to the path

Resolves #1799

### DIFF
--- a/src/generate.js
+++ b/src/generate.js
@@ -43,7 +43,9 @@ define(['lodash'], function(_) {
 
     // Load in all the detects
     _.forEach(config['feature-detects'], function(detect) {
-      detect = detect.indexOf('test/') === 0 ? detect : 'test/' + detect;
+      if (detect.charAt(0) !== '/' && detect.indexOf('test/') !== 0) {
+        detect = 'test/' + detect;
+      }
       output += ', "' + detect + '"';
     });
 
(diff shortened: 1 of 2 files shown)


patrickkettner:
thanks so much for the PR.
I am a bit concerned about a possible security vulnerability from this and the download service that runs on modernizr.com I want to make sure that this can't be exploited there to read the /etc/shadow or  some other non-modernizr file

terinjokes:
@patrickkettner does modernizr.com package in the browser or via a server?

terinjokes:
@patrickkettner absolute paths don't work right now, because of https://github.com/jrburke/requirejs/issues/1461. I am attempting to work around it by changing the logic on this branch to make passed absolute paths relative to the configured `baseUrl`.

As for the modernizr.com version, would the easiest thing be to drop any unknown tests from the query string in the app server that runs it? Where's the source for it?

patrickkettner:
hey @terinjokes sorry for letting this hang around for so long, total failure on my part :[

I am going to close this, as I plan to migrate everything to es6 modules in the very near future. as a result, all of this infra will just be deleted along with it. Thanks again for creating this, and I am sorry I took so long to reply

PR #1811, opened 2015-12-19
Addtest 'bind' in Function.prototype as browser may not support

kocoten1992:


kocoten1992:
Was trying to add 2 file at once, didn't work, i'll add functionprototypebind.js first and update config-all later

PR #1816, opened 2015-12-21
get dest from config

pdot86:
now will look in the modernize-config.json for a “dest” property to
output the modernize.js file in a specific area, just as you can with
the CLI.

### DIFF
--- a/bin/modernizr
+++ b/bin/modernizr
@@ -159,6 +159,9 @@ try {
     if (configPath) {
       config = _.extend(config, require(configPath));
     }
+    if (config && config.dest) {
+      dest = path.join(cwd, config.dest, 'modernizr.js');
+    }
   }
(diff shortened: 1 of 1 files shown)


ryanseddon:
Why do you need this over just running `modernizr -d /path/to/folder`?

pdot86:
I first noticed that this was an option in the Grunt config, so when I tried it using a modernizr-config.json file I saw that it didn't work so I figured I would add it. IMO it also makes it a bit easier to work into a framework or build process using only a config.

ryanseddon:
Fair enough seems reasonable to me. You should also do a PR to the website to add the option to the [docs](https://github.com/Modernizr/modernizr-neue/blob/master/docs/intro.md#configuration-options).

PR #1823, opened 2016-01-05
Add semicolons to avoid unicode feature-detect generating typos

rowanthorpe:
The unicode feature-detect seems to introduce html typos (two html-entities without their trailing semicolons), which caused non-validating code for me.  This PR fixes that.

### DIFF
--- a/feature-detects/unicode.js
+++ b/feature-detects/unicode.js
@@ -26,8 +26,8 @@ define(['Modernizr', 'createElement', 'testStyles', 'isSVG'], function(Modernizr
 
     testStyles('#modernizr{font-family:Arial,sans;font-size:300em;}', function(node) {
 
-      missingGlyph.innerHTML = isSVG ? '\u5987' : '&#5987';
-      star.innerHTML = isSVG ? '\u2606' : '&#9734';
+      missingGlyph.innerHTML = isSVG ? '\u5987' : '&#5987;';
+      star.innerHTML = isSVG ? '\u2606' : '&#9734;';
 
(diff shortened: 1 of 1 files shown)


patrickkettner:
hey @rowanthorpe!
does non validating prevent it from functioning, or is it a conceptual purity thing?

rowanthorpe:
BTW: I am serving my pages as polyglot XHTML5 (serving "application/xhtml+xml" to browsers that can use it) so this particular bug shows up in firebug as a syntax error in the js and as a subsequent well-formedness error in the xhtml. See attached partial-screenshot.

I think the answer to your question is "yes, and yes", because although the unicode-test would still work fine in svg-content (where backslash-notation is used), at a glance the typo would seem to always cause a negative result (often false) to the offsetWidth test in non-svg content (where the html-entities-without-semicolons are used). In xhtml mode the test-characters are apparently not rendered at all due to the well-formedness error, and I guess in non-xml-html they would be rendered as the actual strings "&#5987" and "&#9734" (either way they would be wrong) ...but I am not familiar with the code so take that as the quick impression that it is.

firebug-modernizr-unicode-bug_scrot

patrickkettner:
sorry it took so long to get this merged. I was trying to get xhtml5 testing built into the automated tests, but it isn't working in phantom. For whatever reason, the vml detect just won't work inside of a XHTML5 page, it ends up throwing a `DOM Exception 11`. It does work in chrome, and I ended up finding one other issue that I will be PRing momentarily.

PR #1824, opened 2016-01-05
Make form-validation output valid code

rowanthorpe:
The form-validation code generates html5 code which doesn't validate as xhtml5 (namely,  a missing close-tag and an attribute with no value). This PR fixes that. If you prefer the brevity of vanilla html5 and don't want to cater to the "also valid xml" camp, then feel free to just close this PR (and I will keep it as a custom change for my use, and make a mental note not to submit PRs for any similar changes I might discover in future).

### DIFF
--- a/feature-detects/forms/validation.js
+++ b/feature-detects/forms/validation.js
@@ -41,7 +41,7 @@ define(['Modernizr', 'createElement', 'docElement', 'testStyles'], function(Mode
     // Calling form.submit() doesn't trigger interactive validation,
     // use a submit button instead
     //older opera browsers need a name attribute
-    form.innerHTML = '<input name="modTest" required><button></button>';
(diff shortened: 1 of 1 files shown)


patrickkettner:
Happy to get this in!
However it is not quite valid. You can't nest an a `<button>` inside of an `<input>`. the `<input>` can just be self closing, no?

rowanthorpe:
Ah oops. I was just running on automatic when I added the close-tag and didn't even try to grok the surrounding code's intended actions, but of course \<input\> is a void element [facepalm]. As I presume this has been working fine in non-xml-html5 mode then the xml-equivalent should just be to make the input-tag self-closing (so the \<input\> just sits before the \<button\> rather than around/within/connected-to it, yes?). I will update the PR now with that assumption.

patrickkettner:
cheers!

Issue #1825, opened 2016-01-06
SecurityError: The operation is insecure.


KK-MarcelJahn:
Hi,

we are currently using a custom Build (v. 2.8.3) and also tried the same build based on 3.2 but we are still receiving a SecurityError in Firefox.

This Error is thrown when using Firefox without cookies. So whenever we block Cookies within Firefox the whole application crashes due to this error.

As far as I know Firefox is pretty strict when cookies are blocked. So for example access to localStorage or sessionStorage will return in a SecurityError. So maybe this is the reason for the error we are experiencing....

This is the only detail I can see is:

SecurityError: The operation is insecure.
...{},T.on=function(e,t){this._l[e]||(this._l[e]=[]),this._l[e].push(t),Modernizr.h...

I hope someone can help us.

patrickkettner:
Could you share the build you are using?

KK-MarcelJahn:
Thanks for the quick response!

We are using this build:
http://modernizr.com/download/#-fontface-backgroundsize-borderimage-borderradius-boxshadow-flexbox-hsla-multiplebgs-opacity-rgba-textshadow-cssanimations-csscolumns-generatedcontent-cssgradients-cssreflections-csstransforms-csstransforms3d-csstransitions-applicationcache-canvas-canvastext-draganddrop-hashchange-history-audio-video-indexeddb-input-inputtypes-localstorage-postmessage-sessionstorage-websockets-websqldatabase-webworkers-geolocation-inlinesvg-smil-svg-svgclippaths-touch-webgl-shiv-mq-cssclasses-addtest-prefixed-teststyles-testprop-testallprops-hasevent-prefixes-domprefixes-svg_filters-load

Thank you!

KK-MarcelJahn:
Hi,

After a bit of digging i found that at least the following tests causing issues:

localStorage,
sessionStorage,
indexedDB

I'll remove them from my build to get a stable system, but it would be great if someone could come up with a safe test.

patrickkettner:
is this an error or a console warning?

KK-MarcelJahn:
Unfortunately it's an error and the script is not being executed which is causing a lot of issues.
After removing LocalStorage, SessionStorage and indexedDB everything is working again.

You may want to read this:
http://meyerweb.com/eric/thoughts/2012/04/25/firefox-failing-localstorage/

It's about cookie expiration but it's the same when blocking cookies.

patrickkettner:
[...]

PR #1831, opened 2016-01-13
Fix IndexedDB check in Firefox with disabled cookies

sheff146:
This PR is a fix for issue https://github.com/Modernizr/Modernizr/issues/1825. According to it FF throws a Security Error during IndexedDB check if cookies are disabled.
Simply wrapped these checks in try/catch.

### DIFF
--- a/feature-detects/indexeddb.js
+++ b/feature-detects/indexeddb.js
@@ -16,7 +16,12 @@ define(['Modernizr', 'prefixed'], function(Modernizr, prefixed) {
   // - Firefox shipped moz_indexedDB before FF4b9, but since then has been mozIndexedDB
   // For speed, we don't test the legacy (and beta-only) indexedDB
 
-  var indexeddb = prefixed('indexedDB', window);
+  var indexeddb;
+  try {
+    indexeddb = prefixed('indexedDB', window);
+  } catch (e) {
+  }
+
   Modernizr.addTest('indexeddb', !!indexeddb);
 
   if (!!indexeddb) {
--- a/feature-detects/indexeddbblob.js
+++ b/feature-detects/indexeddbblob.js
@@ -15,11 +15,16 @@ define(['Modernizr', 'addTest', 'prefixed', 'test/indexeddb'], function(Moderniz
 
   Modernizr.addAsyncTest(function() {
     /* jshint -W053 */
-    var indexeddb = prefixed('indexedDB', window);
+    var indexeddb;
     var dbname = 'detect-blob-support';
     var supportsBlob = false;
     var request;
     var db;
+
+    try {
+      indexeddb = prefixed('indexedDB', window);
+    } catch (e) {
+    }
 
     if (!(Modernizr.indexeddb && Modernizr.indexeddb.deleteDatabase)) {
       return false;


patrickkettner:
lgtm! Could you squash down to a single commit? Thanks!

KK-MarcelJahn:
Hi, 

I am afraid this PR will not fix the issue. 
If I'm not mistaken, Modernizr already uses try catch for local and session storage access. Security Errors will not be caught by a simple try catch otherwise we wouldn't experience any issues with local and session storage.

patrickkettner:
@KK-MarcelJahn did you try the fix? Because I am finding the opposite of your claims

image

sheff146:
@KK-MarcelJahn there are no security errors on localStorage and sessionStorage checks in Modernizr v3 unlike v2. The only error is being thrown on IndexedDB check. You can build sources and try.

sheff146:
@patrickkettner squashed.

feature-detects/indexeddb.js
```diff
@@ -16,7 +16,13 @@ define(['Modernizr', 'prefixed'], function(Modernizr, prefixed) {
   // - Firefox shipped moz_indexedDB before FF4b9, but since then has been mozIndexedDB
   // For speed, we don't test the legacy (and beta-only) indexedDB
 
-  var indexeddb = prefixed('indexedDB', window);
+  var indexeddb;
+  try {
+    indexeddb = prefixed('indexedDB', window);
+  } catch (e) {
+    indexeddb = null;
```
patrickkettner: sorry for not catching before, but this line is unnecessary. can you remove it in both files and squash one more time?

sheff146: Yep. Done.

patrickkettner:
cheers - thanks mate

feature-detects/indexeddb.js
```diff
@@ -16,7 +16,12 @@ define(['Modernizr', 'prefixed'], function(Modernizr, prefixed) {
   // - Firefox shipped moz_indexedDB before FF4b9, but since then has been mozIndexedDB
   // For speed, we don't test the legacy (and beta-only) indexedDB
 
-  var indexeddb = prefixed('indexedDB', window);
+  var indexeddb;
+  try {
+    indexeddb = prefixed('indexedDB', window);
+  } catch (e) {
+  }
```
ryanseddon: Put the closing `}` on the same line

ryanseddon: Actually wouldn't we want to return false here since it won't work and to me that's the same as not supported.

feature-detects/indexeddbblob.js
```diff
@@ -15,12 +15,17 @@ define(['Modernizr', 'addTest', 'prefixed', 'test/indexeddb'], function(Moderniz
...
+    var indexeddb;
     var dbname = 'detect-blob-support';
     var supportsBlob = false;
     var request;
     var db;
 
+    try {
+      indexeddb = prefixed('indexedDB', window);
+    } catch (e) {
+    }
```
ryanseddon: Same here return false.

feature-detects/indexeddb.js
patrickkettner: we could return, but we aren't inside of an addTest, so I don't think `return false` would be doing what you think it would be doing

KK-MarcelJahn:
Yeah sorry for the (my) confusion!
This fix is working!

Guess i messed up with the versions i ussed for testing (in prod we are using 2.8).
So we'll update as soon as this fix is merged and a new version is available. Thank you guys!

feature-detects/indexeddb.js
sheff146: @ryanseddon how to do it technically now when PR is closed? Create new PR?

Issue #1834, opened 2016-01-14
Typo in indexeddbblob-test


JaZo:
There is a typo in the indexeddbblob-test which prevents it from running.

In the indexeddb-test 'indexeddb.deletedatabase' is written all lowercase, but the indexeddbblob-test checks for it camelCase 'indexeddb.deleteDatabase'.

In indexeddb-test
```javascript
Modernizr.addTest('indexeddb.deletedatabase', 'deleteDatabase' in indexeddb);
```
In indexeddbblob-test
```javascript
if (!(Modernizr.indexeddb && Modernizr.indexeddb.deleteDatabase)) {
    return false;
}
```

patrickkettner:
fixed in #1835

PR #1835, opened 2016-01-14
Fix indexeddbblob-test (#1834)

JaZo:
Fix for typo in indexeddbblob-test preventing it from running (#1834)


patrickkettner:
thanks!

PR #1849, opened 2016-01-20
Fix case type on line 9

jshen212:
Fixed case type on line 9 to lowercase to match other single-line comments


patrickkettner:
sure, why not

patrickkettner:
thanks a ton for contributing!!

PR #1866, opened 2016-02-02
AMD to ES6 module conversion

ryanseddon:
:skull: DO NOT MERGE :skull: 

I ran our code through the amazing es5to6 codemod and it ran without errors! Seems to of done some funny stuff with our comments but thought I'd push this to discuss at least.


SlexAxton:
woah cool

patrickkettner:
asesome work @ryanseddon! I just updated the PR with some style changes. The next hiccup is istanbul seems to be puking when trying to instrument the code.

If we could bikeshed for a moment...
If we were to  cut a release with just this change, it would technically be a patch release, yes? Since its backwards compatible as far as the API goes and isn't a new feature, per se.

ryanseddon:
Thanks @patrickkettner the codemod seemed to of dropped the trailing new line and duplicated the leading comments, I'm gonna see if I can fix the codemod and run the files again.

patrickkettner:
I actually am mid fixing the newline issue now. Just adding it via sed. will push shortly

patrickkettner:
actually, I did already. in the last commit

ryanseddon:
Yeah i get that but seems the [codemod](https://github.com/nolanlawson/5to6-codemod) itself should be fixed if other people run this on their own project.

patrickkettner:
oh totally, I just meant it wasn't a requirement for the existing code

tribou:
Is this PR still on the radar?  If so, what tasks are still remaining?  I'd love to see Modernizr eventually export modular checks as mentioned in #1204 so that I could pull in individual feature detects without loading the entire library.

e.g.
```js
const passiveCheck = require('modernizr/feature-detects/dom/passiveeventlisteners')
const supportsPassive = passiveCheck()
```

I'm assuming this PR is on the critical path to that.

ryanseddon:
@tribou no specific tasks I just ran this through the codemod, if you can fix the issues with the comment blocks getting munged that would be helpful.

patrickkettner:
I have been messing with this locally, and have it mostly working. There are a few outstanding issues in my mind, however

1. build sizes are a fair bit larger. Because of all of the boilerplate around require statements, a simple build of ~5 detects goes from ~1k gzipped currently, to ~1.8k gzipped. that feels like a hefty price :/. Perhaps some of this could be minimized/codemoded in some way to reduce the size further?
2. I have no idea how to get a webpack based builder working in the browser. I would want the output of a cli build and a website build to be bit for bit identical, but since webpack doesn't provide a browser compatible version, im a bit unsure of what to do

PR #1875, opened 2016-02-12
HbbTV/OIPF video tests for play, pause and seek

philippsimon:
Hi,
I know not many people (beside some developers at European TV channels and some SmartTV developers) could be interested in this test and I also know you can only test it if you can [change the DVB Carousel](http://www.hbbtv-developer.com/site/wiki/index.php/Setting_up_a_HbbTV_environment) of your video signal, but maybe you still want to merge it. At least we at [ProSiebenSat1Digital](http://www.prosiebensat1.com/en/page/home) are really happy to finally test if a device has a working video object.
Best regards,
Philipp


feature-detects/hbbtv.js
```diff
@@ -0,0 +1,14 @@
...
+  "property": "hbbtv",
+  "caniuse": "hbbtv",
+  "tags": ["hbbtv"]
+}
+!*/
+/* DOC
+TODO
+*/
+define(['Modernizr'], function(Modernizr) {
+  Modernizr.addTest('hbbtv', !!navigator.userAgent.match(/(HbbTV|Opera TV)/i));
```
patrickkettner: this is not a feature detect

feature-detects/hbbtv/video.js
```diff
@@ -0,0 +1,524 @@
...
+/* DOC
+Tests if the play, pause and seek functionality of the HbbTV video object works correctly.
+
+At the moment the [test video](https://cdn.rawgit.com/philippsimon/Modernizr/hbbtv/feature-detects/hbbtv/test.mp4) is loaded from Github.
+For a reliable use I suggest to host the video yourself and set `window.MODERNIZR_HBBTVVIDEO_TEST_VIDEO` to that URL.
+*/
+define(['Modernizr', 'addTest', 'getBody', 'createElement', 'test/hbbtv'], function(Modernizr, addTest, getBody, createElement) {
+
+  /*
+   * Leave the console statements inside as it's the only good working way to debug the problematic HbbTV devices
```
patrickkettner: nah

feature-detects/hbbtv/video.js
```diff
@@ -0,0 +1,524 @@
...
+        //initApp();
+
+        var videoId = video.id;
+
+        var tests = {
+
+          play: function() {
+            try {
+              video.play(1);
+              // console.log('video.play');
```
patrickkettner: we need to delete any commented out code

feature-detects/hbbtv/video.js
```diff
@@ -0,0 +1,524 @@
+/* global setTimeout */
+/*!
+{
+  "name": "HbbTV Video",
+  "property": "hbbtvvideo",
+  "tags": ["hbbtv", "oipf", "video"],
+  "async" : true,
+  "warnings": [
+    "This test is currently only compatible to HbbTV >= 1.x",
+    "All siblings of an existing HbbTV video object in the DOM will be deleted during the test"
```
patrickkettner: this can't happen.

patrickkettner:
I would love to have this merged, but there is a number of issues with it in its current state. I added a few comments to begin with, but in addition 

- we can't have an external video download needed (i.e. the mp4 file). Make it as small as possible then base64 to include it.
- This test is HUGE. We will need to greatly reduce the amount of excess code in there to get it in.
- We shouldn't be committing commented out code. till only confuse folks in the future.


I'd suggest checking out something like the video autoplay test and basing this more off of that.

please don't hesitate to say something if you have an questions, and thanks again!

feature-detects/hbbtv/video.js
philippsimon: The problem is that HbbTV < 2.x only allows one video object to be inside the DOM and so I have to remove the old video object.
Do you think it would be ok to tell the user that the test requires that there is no other video object in the DOM during the test? It would reduce the amount of code a lot.

feature-detects/hbbtv/video.js
philippsimon: So I would prefer to leave the comments inside as they are the only way to debug the devices (for testing I replaced the console with a function that displays it on the device). I tested the code on around 25 TVs and half of them were extremely buggy and crashed at really weird places (worse then an old Internet Explorer ;) ).

But I could also leave them just inside of my repository and make a pull request with a clean branch.

philippsimon:
Hi Patrick,
thanks for your comments!
Here my reasons why I wrote my code like this:
- I based my code on the autoplay test but sadly base64 encoded videos in the data field are not supported by several HbbTV devices.
- Yes - it's way to big. One big problem is that some devices replace the video object during the initialization with a new object and copy the properties. And also I need to put timeouts and intervals at all kind of parts as this initialization process needs sometimes a lot of time and if I access a property before the device just crashes. One way to reduce the code a lot would be to require that there is no other video object in the DOM during the test. Do you think that's fine?

And believe me - I also hate it that the TV manufactures don't care at all to implement a working browser and even more that none gives you access to the debugging console ...

feature-detects/hbbtv/video.js
patrickkettner: What happen if there is more than one video?

you should be able to `cloneNode()` on the video element, cache that, and then readd it later, no?

feature-detects/hbbtv/video.js
patrickkettner: totally get that you need to debug while writing the script (I do similiar things all the time), but those downloading the script expect it to fully work and it should not need any kind of debugging.

patrickkettner:
> I based my code on the autoplay test but sadly base64 encoded videos in the data field are not supported by several HbbTV devices.
Unfortunately we won't be able to merge it in then :[

It really does look like you've done some valiant work here, and all the more so because it is a little developed field for obscure browsers (I can only imagine how much fun that was), but it is verboten to have external assets required. Perhaps you could have this on a gist or on your own repo as a Modenrizr plugin?

feature-detects/hbbtv.js
philippsimon: ok I remove that

philippsimon:
Yes - soo much fun ... only 2 weeks full time ...
And can this plugin then get listed on the Modernizr page?
Would be just nice to push this kind of tests more so the TV producers get forced to develop good working devices.

patrickkettner:
> Yes - soo much fun ... only 2 weeks full time ...

my condolences :[

> And can this plugin then get listed on the Modernizr page?

I'd be happy to have a post on [the blog](https://modernizr.com/news) about it

> Would be just nice to push this kind of tests more so the TV producers get forced to develop good working devices.

Certainly, and speaking as an employee of a browser, please make sure to try and file bugs whenever you find them - it REALLY helps prioritize things when folks tell us how we screwed up (especially with code!)

philippsimon:
Thanks - that with the blog sounds great! I will let you know when I finished the plugin.

Sadly I guess the big TV producers don't care much about the (HbbTV) browser. It seems they want to push more their own SmartTV app stores and offer the other (normally very outdated) browser(s) only to mention it in the advertisement. When I use the current version of the Opera browser in VirtualBox, which is used on many devices, these problems never happen.

philippsimon:
Hey @patrickkettner, we created a repository for the HbbTV plugins. As I couldn't add them with the `modernizr-config.json` the user has to append them to the Modernizr file.
Or do you know a better way to get one Modernizr file with the selected tests that are relevant for HbbTV plus our own plugins?

Here's our repository. At the moment the ReadMe is still quite basic, so I will let you know when we wrote something nice and it would be great if you could then mention it on the Modernizr page:
https://github.com/p7s1digital/modernizr-hbbtv-plugins

PR #1883, opened 2016-02-16
Fix testProps so it works on strict XHTML browsers

philippsimon:
Without `elems.length` it will be an endless loop if the browser can't create the element.
For strict XHTML browsers I added the hardly used element `samp` to test the style. Samsung SmartTVs from 2011 end up otherwise in the endless loop.

### DIFF
--- a/src/testProps.js
+++ b/src/testProps.js
@@ -32,8 +32,9 @@ define(['contains', 'mStyle', 'createElement', 'nativeTestProps', 'is', 'cssToDO
     // inside of an SVG element, in certain browsers, the `style` element is only
     // defined for valid tags. Therefore, if `modernizr` does not have one, we
     // fall back to a less used element and hope for the best.
-    var elems = ['modernizr', 'tspan'];
-    while (!mStyle.style) {
+    // for strict XHTML browsers the hardly used samp element is used
+    var elems = ['modernizr', 'tspan', 'samp'];
+    while (!mStyle.style && elems.length) {
       afterInit = true;
       mStyle.modElem = createElement(elems.shift());
       mStyle.style = mStyle.modElem.style;


philippsimon:
Can somebody rerun the test? I tested it locally and I don't get any timeouts.

patrickkettner:
which tests had `elems.length === 0`? This is covering up a bug, rather than fixing it

philippsimon:
CSS animations didn't work on a Samsung 2011 TV which accepts only valid XHTML elements which neither `modernizr` nor `tspan` is. With `samp` it passed the test.

And I would leave the `elems.length` inside just to be sure that it really crashes the next time and doesn't stay in the loop.

patrickkettner:
My point is  that we need to fix the feature detect to make that not happen, not this

philippsimon:
I checked the feature detect and it's fine:
The browser doesn't return for `createElement('modernizr' or 'tspan')` a correct HTML element as it's a XHTML document and both are not in the list of allowed elements:
https://www.w3.org/TR/2005/WD-xhtml2-20050527/elements.html

patrickkettner:
if it passing a length of 0 to this helper, it is not fine

philippsimon:
No - just the while loop is shifting the 2 lines above declared elems array and can't create any of the 2 elements as they are just invalid in an XHTML document (yes - most browsers accept it, but actually they shouldn't allow it).

patrickkettner:
If they are invalid in an XHTML document then they need to be made valid. Change the element to something that is valid

philippsimon:
The problem is that an element is needed that is valid in `XHTML`, `HTML4`, `HTML5` and `SVG`.

So for sure for HTML5 is `modernizr` the most secure one and valid.

For the others a valid element should be used that is hardly.

For `HTML4` and `XHTML` I would suggest `samp` as till today I didn't even know that it exists. Here the lists of allowed elements:
https://www.w3.org/TR/2005/WD-xhtml2-20050527/elements.html
https://www.w3.org/TR/html4/index/elements.html

As I don't have experience with `SVG` I would suggest to just use the suggested `tspan` (sadly it's not a valid `HTML` element).

So my conclusion is: As that is what my code actually does just merge my branch ;)

The `elems.length` could be also removed I just want to be sure that the test really crashes if the browser doesn't accept any of the elements. It's much faster to find then the bug instead of adding `console.log` statements everywhere.

patrickkettner:
The Angie tag works everywhere, I believe. I'll have to test

philippsimon:
I know - actually that TV is the only one which can't create the element and has this weird browser. The browser seemed to have implemented only the absolutely required parts and supports only XHTML 1.0 Strict. The device uses one of the first versions of Samsungs self developed browser Maple and it seems it doesn't support any HTML5:
http://stackoverflow.com/questions/8290609/how-to-test-web-development-for-the-maple-browser-installed-on-samsung-tvs

patrickkettner:
sorry, I meant anchor, not Angle. autocorrect sucks :[

What elements DOES it support?

On Wed, Feb 17, 2016 at 12:29 AM, Philipp Simon Schmidt <
notifications@github.com> wrote:

> I know - actually that TV is the only one which can't create the element
> and has this weird browser. The browser seemed to have implemented only the
> absolutely required parts and supports only XHTML 1.0 Strict. The device
> uses one of the first versions of Samsungs self developed browser Maple and
> it seems it doesn't support any HTML5:
>
> http://stackoverflow.com/questions/8290609/how-to-test-web-development-for-the-maple-browser-installed-on-samsung-tvs
>
> —
> <https://github.com/Modernizr/Modernizr/pull/1883#issuecomment-185092887>.
>



-- 
patrick

philippsimon:
I just see your point: So it supports the anchor tag but as I understood the testProps function doesn't want to create any anchor tag. I take a look at it and write you later.

patrickkettner:
@philipbordallo hows it goin?

philippsimon:
Hi @patrickkettner. I'll take a look at it this week. The last weeks were really busy.

patrickkettner:
no worries! take yer time. just checking :]

On Tue, Mar 8, 2016 at 12:17 AM, Philipp Simon Schmidt <
notifications@github.com> wrote:

> Hi @patrickkettner <https://github.com/patrickkettner>. I'll take a look
> at it this week. The last weeks were really busy.
>
> —
> <https://github.com/Modernizr/Modernizr/pull/1883#issuecomment-193657164>.
>



-- 
patrick

philippsimon:
ok - so I went through the code and got quite irritated by this:
https://github.com/Modernizr/Modernizr/blob/master/src/testAllProps.js
`ModernizrProto.testAllProps = testAllProps;`
https://github.com/Modernizr/Modernizr/blob/master/src/testPropsAll.js
`ModernizrProto.testAllProps = testPropsAll;`
(i guess it should be `ModernizrProto.testPropsAll = testPropsAll`)

but in general my fix is fine:
i.e. `testAllProps('animationName', 'a', true)` creates now one of the 3 element types (`modernizr` for `HTML5`, `tspan` for SVG, `samp` for `XHTML`) and sets then the style property `animationName` to `a`. So the problem was really only that it only tried to create a `modernizr` or `tspan` element which is forbidden by the `XHTML` specifications.

The `elems.length` I would still leave inside as maybe there are browsers that also forbids these 3 element types (I just hope not ...).

patrickkettner:
> So the problem was really only that it only tried to create a modernizr or tspan element which is forbidden by the XHTML specifications
Why is this a problem? in a practical sense, I mean

philippsimon:
So the HbbTV browsers of the Samsung TVs from 2011 (a highly modified Gecko engine) returns for `createElement` of the type `modernizr` or `tspan` an object which doesn't have the style property.

patrickkettner:
Sorry, perhaps I am jsut being thick, but does your update mean that it now functions on those HbbTV browsers?

philippsimon:
yes ;)

philippsimon:
Hi @patrickkettner, can please you merge my commit?

ryanseddon:
@patrickkettner I'm :+1: on this/

patrickkettner:
Sorry for taking so long to merge! looks great

PR #1884, opened 2016-02-16
Fix websockets test for some Panasonic SmartTVs

philippsimon:
SmartTV browsers from 2012 Panasonic TVs crash otherwise.
It seems they didn't integrate `window.WebSocket` fully.

### DIFF
--- a/feature-detects/websockets.js
+++ b/feature-detects/websockets.js
@@ -25,5 +25,9 @@
 }
 !*/
 define(['Modernizr'], function(Modernizr) {
-  Modernizr.addTest('websockets', 'WebSocket' in window && window.WebSocket.CLOSING === 2);
+  var supports = false;
+  try {
+    supports = 'WebSocket' in window && window.WebSocket.CLOSING === 2;
+  } catch (e) {}
+  Modernizr.addTest('websockets', supports);
 });


philippsimon:
Can somebody rerun the test? I tested it locally and I don't get any timeouts.

feature-detects/websockets.js
```diff
@@ -25,5 +25,9 @@
 }
 !*/
 define(['Modernizr'], function(Modernizr) {
-  Modernizr.addTest('websockets', 'WebSocket' in window && window.WebSocket.CLOSING === 2);
+  try {
+    Modernizr.addTest('websockets', 'WebSocket' in window && window.WebSocket.CLOSING === 2);
+  } catch (e) {
+    Modernizr.addTest('websockets', false);
```
patrickkettner: Rather than running `addTest` twice, use this pattern

```
var supports;

try {
  supports = blah in window
} catch (e) {
  supports = false
}

Modernizr.addTest('blah', supports);
```

philippsimon: I would say it wouldn't even run the test, but yes it's more clear

patrickkettner: > I would say it wouldn't even run the test
pardon?

philippsimon:
... appveyor seems to be really slow. I would suggest to increase the timeout.

philippsimon:
@patrickkettner as my pull request to increase the timeout was merged: Could you please rerun the appveyor test and then merge? Thanks!

feature-detects/websockets.js
```diff
@@ -25,5 +25,11 @@
 }
 !*/
 define(['Modernizr'], function(Modernizr) {
-  Modernizr.addTest('websockets', 'WebSocket' in window && window.WebSocket.CLOSING === 2);
+  var supports;
+  try {
+    supports = 'WebSocket' in window && window.WebSocket.CLOSING === 2;
+  } catch (e) {
+    supports = false;
```
patrickkettner: You don't need to rest it here. Just set the initial variable value to false, and do an empty catch. Can you update and squash?

philippsimon:
@patrickkettner: updated and squashed

PR #1889, opened 2016-02-17
Increase mochaTest timeout to 5s

philippsimon:
As the mocha tests on appveyor failed on many pull requests because of timeouts I increased the timeout for the mochaTest to 5 seconds.


patrickkettner:
Thanks!

PR #1919, opened 2016-03-09
Passthrough script for Powershell and cmd.exe

roblarsen:
This batch files catches calls to the modenizr script, tells the
environment to use node and then passes along all arguments.

Tested on Windows 10 in Powershell,cmd.exe and Git Bash


ryanseddon:
Thanks @roblarsen

PR #1920, opened 2016-03-09
Fix preserve3d's IE11/Win10 false positive

### LINKED ISSUE #1748: Modernizr.preserve3d returns false positive in IE11 on Windows 10
Modernizr.preserve3d returns true in IE11 on Windows 10 (incorrect).
In IE11 on Windows 7 it returns false (correct).

Found this issue in Modernizr 3.2.0.

paulirish:
This is @afarkas's patch. I'm just helping :smiley_cat: 

fixes #1748


paulirish:
test page: https://output.jsbin.com/bajubol/1/quiet

win7 ie11 - should be false
image

win10 ie11 - should be true
image

lookin good.

ryanseddon:
Is't this the `preserve3d` test looks false on both in your screenshot still?

paulirish:
> Isn't this the `preserve3d` test looks false on both in your screenshot still?

it does appear that is true... :/

ryanseddon:
@aFarkas is this test missing anything? Seem to be still getting a false negative in win10?

aFarkas:
Not sure, but I think your expectations are wrong. IE11 should always return `false` (in both cases).

The problem was, that IE11 on Win10 returned `true` although it should return `false`. Now it returns always `false`. 

This is also what the title says.

@paulirish Thank you very much to add the modified test (it's so important to avoid layout in modern browsers.)

ryanseddon:
so if we rebase or merge in master this should be god to go right @paulirish @aFarkas?

paulirish:
Rebased and updated.  @aFarkas can you review my logic flip in f9ddfa2 ?

aFarkas:
LGTM

ryanseddon:
Build failing was a linting issue merged to master so this is good to go, merging.

paulirish:
Ace. Thanks for the followup, @ryanseddon !

PR #1925, opened 2016-03-15
fixes Edge regex in uaparser.js

### LINKED ISSUE #1902: CanIUse tests fail in Edge
In the Modernizr Master branch 80 CanIUse tests are failing on Edge.  This doesn't repro on v3.3.1 of Modernizr, only the master branch,

Most of the failures mention IE5.5 
`IE5.5: Caniuse result for download matches Modernizr's result for adownload `

Possibly a bad ua sniff?

I have been trying to root cause the issue but I think I still have a lot of distance to cover before I understand how these tests work

AmazingJaze:
fixes #1902 

This same fix has also already been applied at [uap-core](https://github.com/ua-parser/uap-core/commit/87b6282f3e2792fc0bfb89f4a84397417329dbd0)

Without this change Modernizr caniuse tests incorrectly identify Edge as IE 5.5 and 80 of them fail.


AmazingJaze:
//CC @patrickkettner @jdalton

AmazingJaze:
@paulirish Any chance we can get this merged? This is a very simple change that will fix 80 failing Modernizr tests in Edge.

jdalton:
:sunglasses:

paulirish:
JD, you're a chill son of a gun.

PR #1947, opened 2016-04-05
Prefixed returns a bound constructor

arlolra:
In testDOMProps.js, this doesn't always seem like the best thing to do,

```
// let's bind a function
if (is(item, 'function')) {
  // bind to obj unless overriden
  return fnBind(item, elem || obj);
}
```

### DIFF
--- a/feature-detects/webrtc/datachannel.js
+++ b/feature-detects/webrtc/datachannel.js
@@ -11,24 +11,17 @@
 /* DOC
 Detect for the RTCDataChannel API that allows for transfer data directly from one peer to another
 */
-define(['Modernizr', 'prefixed', 'domPrefixes', 'test/webrtc/peerconnection'], function(Modernizr, prefixed, domPrefixes) {
-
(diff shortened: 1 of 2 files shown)


patrickkettner:
Can you please explain your reasoning more?

arlolra:
`window.RTCPeerConnection` is a constructor.

When I do `prefixed('RTCPeerConnection', window)`, I'm returned a bound function, which prevents me from constructing a `new` object. In other words, `prefixed` assumes I'm going to be invoking the returned function.

arlolra:
Sorry, I should add, the commits here are just cleanup / simplification.

The subject of the PR has to do more with the comment in there, highlighting the false assumption.

arlolra:
Pushed another commit to explain further.

ryanseddon:
This would be a breaking change for people relying on the autobound nature of `prefixed`, I don't have any data on who is using it and if they rely on the autobound nature. 

Perhaps an escape hatch for prefixed to not bind via an argument would be better.

PR #1957, opened 2016-04-12
Fixes Preserve3D test in Edge and IE. 

### LINKED ISSUE #1900: "svg context: is able to be loaded in a SVG file" test is failing in Edge
100% repro in the master branch of Modernizr on Edge. However, doesn't repro in [v3.3.1](https://github.com/Modernizr/Modernizr/tree/v3.3.1)

`Error: expected 'Unable to set property \'cssText\' of undefined or null reference http://localhost/TestContent/Modernizr/master/dist/modernizr-build.js 4478 5' to equal undefined
at Assertion.prototype.assert (http://localhost/TestContent/Modernizr/master/node_modules/expect.js/index.js:102:7)
   at Assertion.prototype.equal (http://localhost/TestContent/Modernizr/master/node_modules/expect.js/index.js:216:5)
   at Anonymous function[name] (http://localhost/TestContent/Modernizr/master/node_modules/expect.js/index.js:69:13)
   at object.onerror (http://localhost/TestContent/Modernizr/master/test/browser/integration/svg.js:19:13)
`

I can see that a change was made in feature-detects/css/transformstylepreserve3d.js that is setting CSSText

https://github.com/Modernizr/Modernizr/compare/v3.3.1...master

AmazingJaze:
Closes #1900

### DIFF
--- a/feature-detects/css/transformstylepreserve3d.js
+++ b/feature-detects/css/transformstylepreserve3d.js
@@ -21,8 +21,8 @@ define(['Modernizr', 'createElement', 'docElement'], function(Modernizr, createE
     var outerAnchor = createElement('a');
     var innerAnchor = createElement('a');
 
-    outerAnchor.style.cssText = 'transform-style: preserve-3d; transform-origin: right; transform: rotateY(40deg);';
-    innerAnchor.style.cssText = 'width: 9px; height: 1px; background: #000; transform-origin: right; transform: rotateY(40deg);';
+    outerAnchor.style.cssText = 'display: block; transform-style: preserve-3d; transform-origin: right; transform: rotateY(40deg);';
+    innerAnchor.style.cssText = 'display: block; width: 9px; height: 1px; background: #000; transform-origin: right; transform: rotateY(40deg);';
 
     outerAnchor.appendChild(innerAnchor);
     docElement.appendChild(outerAnchor);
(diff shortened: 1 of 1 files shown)


aFarkas:
@AmazingJaze 
What was wrong with [this PR](https://github.com/Modernizr/Modernizr/pull/1920)?

AmazingJaze:
It results in an exception being thrown in Modernizr's svg context test when run in Edge and IE.
See this entry from the linked issue:
https://github.com/Modernizr/Modernizr/issues/1900#issuecomment-188986911

AmazingJaze:
updated inline^ for clarification

AmazingJaze:
@aFarkas, my mistake I thought the PR you referenced was the PR that originally introcuded the exception reported by #1900. If #1920 already addresses the failure caused by #1900 then this PR is no longer necessary. @aFarkas can you confirm?

aFarkas:
@AmazingJaze 
Thx, but I was also mixing up things. Your PR is still necessary. My hope was only that we could somehow guard the test using `CSS.supports`. (Therefore #1920 would only help in EDGE not in IE.)

The preserve3D introduces a write -> read -> write sequence and it would be great to only do this, if we absolutely need to. Maybe merge the `CSS.supports` part with your fix?

PR #1963, opened 2016-04-25
Improve videoautoplay detection

mfairchild365:
videoautoplay detection can be flaky. 300ms is a long enough wait for many sites, however there are plenty of other sites where that number is too low. Rather than simply increasing the wait, this commit tries to address the problem by increasing the wait time to 1000ms while re-checking every 200ms to ensure the quickest result. While this might not be enough for every situation, it should cover many more situations.


tmsdnl:
Any chance this is getting to master any time soon?

ryanseddon:
Thanks @mfairchild365 we'll see how this goes.

tmsdnl:
@ryanseddon is there a plan to bump version as well?

PR #1964, opened 2016-04-26
Remove duplicate img/jpeg2000 feature-detects entry

scottaddie:
This PR removes a duplicate `img/jpeg2000` entry from the `feature-detects` array within the `lib/config-all.json` file.


patrickkettner:
cheers!

Issue #1789, opened 2015-12-04
Modernizr v3 flash detect always returning {block: true} on Mac Safari 9.0 and Windows 10 Edge


nireshkumar:
On Mac Safari v9 and Windows 10 Edge browser flash detects always returns {block:true}, works fine on other browsers.

patrickkettner:
I can't replicate this on either Edge or Safari 9. Could you share any more information or example?

nireshkumar:
Here is the screenshot and browser info for safari and Edge

Browser info - Windows 10 - Edge : http://whatsmybrowser.org/b/LI7MOCI
![Windows 10 - Edge](https://cloud.githubusercontent.com/assets/1169368/11621027/1f55f150-9cde-11e5-9493-889bb08deaef.PNG)

Browser info - OS X El Capitan - Safari : http://whatsmybrowser.org/b/HPUVDMX
![OS X El Capitan - Safari](https://cloud.githubusercontent.com/assets/1169368/11621036/30ea62ac-9cde-11e5-8691-4bf51d78429c.png)

Please let me know if you need more info or some other ways to reproduce the issue

nireshkumar:
@patrickkettner Let me know if you need any other details to debug the issue

Thanks

geoff346:
I've experienced identical issues. I haven't been able to test on the latest mac, but I get an identical false positive on microsoft edge with windows 10.

I can't see where it's breaking at the moment, but if I turn flash OFF in edge it correct reports flash isn't installed.
If I turn Flash back on, it incorrectly reports flash on, but BLOCKED.
My system is a recent upgrade from win7 to 10: 

Testing with modernizer on my own custom page hasn't produced any better results, but I did notice that putting the modernizr script in the html head of the page also breaks flash test with firefox and chrome (I haven't flagged that up separately as it's fixed for firefox and chrome by putting modernizr script tag into the html body). If it is related, the flash test might need to wait for body load before firing?

FullUserAgent=
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2486.0 Safari/537.36 Edge/13.10586
Flash Player 20

cjanik:
I've just encountered this issue, and have submitted a PR to address it.
https://github.com/Modernizr/Modernizr/pull/1972

PR #1972, opened 2016-05-04
attach body once before appending embed

cjanik:
When (fake) body is not attached before appending embed, detection will falsely report that blocking is active. The attachment only needs to be done once, as body will still be attached if previous check is inconclusive.

### DIFF
--- a/feature-detects/flash.js
+++ b/feature-detects/flash.js
@@ -68,12 +68,12 @@ define(['Modernizr', 'createElement', 'docElement', 'addTest', 'getBody', 'isSVG
       embed.type = 'application/x-shockwave-flash';
 
       // Need to do this in the body (fake or otherwise) otherwise IE8 complains
+      attachBody(body);
       body.appendChild(embed);
 
       // Pan doesn't exist in the embed if its IE (its on the ActiveXObjeect)
       // so this check is for all other browsers.
       if (!('Pan' in embed) && !activex) {
-        attachBody(body);
         runTest('blocked', embed);
         removeFakeBody(body);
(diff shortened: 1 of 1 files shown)


patrickkettner:
hi @cjanik!
Thanks for the PR - I am not quite clear on how the bug was manifesting. Could you explain further, or make a quick repro?

thanks!

patrickkettner:
ping @cjanik! I'd like to ensure we don't regress this in the future, so I would love to know how to replicate it

patrickkettner:
hey @cejanik @cjanik - sup :]

jessemorton:
@patrickkettner is the thread at #1789 enough context to judge? There's just unreliable results in some different browsers. I haven't tested @cjanik's fix myself, but can confirm some unreliability in Flash testing. Chrome 56 on Mac is always telling me it's blocked, for example.

dbeg:
@patrickkettner @jessemorton 

I'm seeing what appears to be the same false positive blocking issue that this case might be addressing (as well as what was referenced within #1789):

When Flash is installed and enabled, `Modernizr.flash` incorrectly returns `{ blocked: true }` (the same result occurs when using `Modernizr.on('flash'...`.

I was able to reproduce with the following browers:

- [Chrome 57 on OS X 10.11](http://whatsmybrowser.org/b/WI9YOJ6)
- [FF 53 on OS X 10.11](http://whatsmybrowser.org/b/PHE9C3S)

screen shot 2017-05-01 at 4 44 20 pm

screen shot 2017-05-01 at 4 42 58 pm

Issue #1365, opened 2014-06-13
css-scrollbars false positive in firefox


codekipple:
Seems to be returning true in firefox but as far as i know there is no way to styles scrollbars in the latest version of firefox.

patrickkettner:
ping @codekipple

codekipple:
@patrickkettner Hi, yeah ok I'll get on it and create a pull request. It's been a while since I tried the css-scrollbars test. I'll fire it up and re test it before submitting pull request.

patrickkettner:
:D thanks dude - don't hesitate to ask for a hand, would be stoked to have you land it

patrickkettner:
@codekipple hows it goin?

galtalmor:
I think this should be closed now...

PR #1990, opened 2016-05-22
Fix issue #1365 - css-scrollbars false positive in FF

galtalmor:
Applied a fix according to the conversation in issue #1365 .


patrickkettner:
thanks!

PR #1993, opened 2016-05-23
localstorage: add compat notes

paulirish:
Adding some extra compatibility notes around higher-security settings.


patrickkettner:
thanks paul!

PR #2044, opened 2016-07-31
convert jshint and jscs to eslint

patrickkettner:
because im hipster as hell

also, there are a number of places custom rules would be useful, so moving to
eslint makes the creation thereof easier


alrra:
> because im hipster as hell

Also, [_"JSCS is now considered fully deprecated."_](http://eslint.org/blog/2016/07/jscs-end-of-life).

patrickkettner:
> Also, "JSCS is now considered fully deprecated.".

Having worked in corporate environments for quite a while now, if you aren't running something that is deprecated in production, you are a hpister

patrickkettner:
ping @ryanseddon just a heads up this happened :]

PR #2063, opened 2016-08-31
Fix "elem.parentNode is null" with Flashblock

jdelStrother:
The Flashblock Firefox extension seems to take the questionable approach of removing `video` nodes immediately as they're inserted into the page, which causes an "elem.parentNode is null" error.  What do you think to this fix?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -46,7 +46,7 @@
       "error",
       "single"
     ],
-    "curly": "off",
+    "curly": "error",
     "wrap-iife": "off",
     "no-caller": "error",
     "no-cond-assign": [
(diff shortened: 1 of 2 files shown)


patrickkettner:
hey @jdelStrother !
That looks great, could you change it to have smoe curly braces and break it up over a couple lines?

thanks!

jdelStrother:
Sure, I'll do so tomorrow.   (Though if that is the preferred style, shouldn't eslint specify `"curly": "all"` https://github.com/Modernizr/Modernizr/blob/10296a8215cee98d30e3f084ec3f79c8b4d828f8/.eslintrc#L49 ?)

ryanseddon:
@jdelStrother good idea want to sneak in the eslint change in this PR and see if our linting still passes if not we can do a follow up PR.

jdelStrother:
Done.   Looks like the rest of the code is already using braces, so I've included the .eslintrc change in this PR.   I can break it out if it proves contentious.

patrickkettner:
ah man, _perfect_ use of commits :]

patrickkettner:
thanks so much!

PR #2066, opened 2016-09-05
Remove duplicates from config-all.json

camjc:
Unless there is a good reason for these to be here?


patrickkettner:
nope!

thanks very much :D

PR #2091, opened 2016-10-04
Repair detect regions

### LINKED ISSUE #2079: Modernizr.prefixed is not a function error when including CSS Regions test
Using a custom build of Modernizr v3.3.1, I get an "Uncaught TypeError: Modernizr.prefixed is not a function" only when including the "CSS Regions" test. If I exclude that test in the build, I get no errors in the console.

The build is generated by gulp-modernizr, and I can confirm that the Node package for Modernizr is up-to-date (3.3.1). I can also confirm the error occurs when replacing the custom build with a build from modernizr.com containing only the "CSS Regions" test.

Possibly related to this issue: https://github.com/Modernizr/Modernizr/issues/948

Isthu:
Detect regions is broken badly. It invokes an undefined property
causing failure to run _any_ detects. The cause is a missing
dependency.

Fixes this by including the missing dependency. Also updates the
references to the free variable style, the correct pattern
in the AMD module architecture.

Fixes #2079

### DIFF
--- a/feature-detects/css/regions.js
+++ b/feature-detects/css/regions.js
@@ -12,7 +12,7 @@
   }]
 }
 !*/
-define(['Modernizr', 'createElement', 'docElement', 'isSVG'], function(Modernizr, createElement, docElement, isSVG) {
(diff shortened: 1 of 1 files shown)


Isthu:
Test of a build with this fix installed. The detect is executed and the outcome is displayed:
http://codepen.io/Isthu/pen/rrYvkB

PR #2095, opened 2016-10-06
Remove blacklisting of old Android from fontface.js. Closes #1851

### LINKED ISSUE #1851: False negative for @font-face on Android Firefox
I have a false negative detection of @font-face support on Android 4.4.2 with Firefox 43.0 in latest modernizr (3.3.1).

The `blacklist` function returns true for this combination as `wkvers` is `null` and `ua` match Android regexp. In previous version of modernizr (2.8) the detection was fine.

PikadudeNo1:
Usage of pre-2.2 Android versions has fallen below 0.1% [(source)](https://developer.android.com/about/dashboards/index.html), so @font-face is practically guaranteed to work on Android. The blacklisting was catching all versions of Firefox for Android.

### DIFF
--- a/feature-detects/css/fontface.js
+++ b/feature-detects/css/fontface.js
@@ -29,11 +29,9 @@
 define(['Modernizr', 'testStyles'], function(Modernizr, testStyles) {
   var blacklist = (function() {
     var ua = navigator.userAgent;
-    var wkvers = ua.match(/applewebkit\/([0-9]+)/gi) && parseFloat(RegExp.$1);
     var webos = ua.match(/w(eb)?osbrowser/gi);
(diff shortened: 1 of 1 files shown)


patrickkettner:
honestly, i'd rather fix the ff issue than remove the couple bits of code.

@ryanseddon @SlexAxton thoughts?

PikadudeNo1:
No comments for 10 weeks. Issue #1851 continues to affect Firefox for Android users. What's next?

I took the approach I did because I think even the guarantee of false positives on Android <2.2 is less important than the chance of another UA-sniffing bug in the future. After all, avoiding UA-sniffing bugs is part of why developers choose Modernizr over UA-sniffing.

PR #2136, opened 2016-12-28
contains => includes (String.prototype)

philippevk:
See https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/includes


ryanseddon:
oh good catch, thanks.

PR #2164, opened 2017-02-22
New Feature detection: formNoValidate

andreacc:
Feature detection for formNoValidate attribute on inputs.


rejas:
any reason for not merging this @patrickkettner (besides that only very old browsers would fail it)?

patrickkettner:
thanks!

PR #2177, opened 2017-03-16
Fix rgba/hsla false negative in browsers that alias them to rgb/hsl

PikadudeNo1:
Fixes incompatibility with the newly-released Firefox 52. I decided to do nothing about the pre-existing inconsistent (non-)use of `contains`.

### DIFF
--- a/feature-detects/css/hsla.js
+++ b/feature-detects/css/hsla.js
@@ -10,6 +10,6 @@ define(['Modernizr', 'createElement', 'contains'], function(Modernizr, createEle
   Modernizr.addTest('hsla', function() {
     var style = createElement('a').style;
     style.cssText = 'background-color:hsla(120,40%,100%,.5)';
-    return contains(style.backgroundColor, 'rgba') || contains(style.backgroundColor, 'hsla');
+    return contains(style.backgroundColor, 'rgb') || contains(style.backgroundColor, 'hsl');
   });
 });
--- a/feature-detects/css/rgba.js
+++ b/feature-detects/css/rgba.js
@@ -15,6 +15,6 @@ define(['Modernizr', 'createElement'], function(Modernizr, createElement) {
     var style = createElement('a').style;
     style.cssText = 'background-color:rgba(150,255,150,.5)';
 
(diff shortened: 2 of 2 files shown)


patrickkettner:
hey @PikadudeNo1! 
Thanks so much for reporting. without more information, I feel like this may be an accidental change in firefox as `getComputedStyle` still returns the `rgba` value, and the parsed value is not a valid `rgb` value (unless something has changed at the spec level recently that I am not aware of).

I reached out to @miketaylr on twitter to see if he has any background

PikadudeNo1:
There has indeed been a spec-level change; [the latest CSS Color draft](https://drafts.csswg.org/css-color/) lets rgb and hsl take an optional alpha value and has rgba and hsla as legacy aliases. [Mozilla's color value documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) says Firefox 52 is the first version of Firefox to support this change.

patrickkettner:
in speaking with some of the CSS spec folks I work with on Edge, they also feel that the value here should not be what firefox is returning. I would not be surprised if we did merge this, but I would like to get feedback from engineers at mozilla before doing so

patrickkettner:
yeah, confirmed by @dholbert as incorrect behavior.

image

this has [already been changed](https://bugzilla.mozilla.org/show_bug.cgi?id=1347164) in their code. as a result, I don't think it makes sense to merge this change.

Despite that, I _really_ appreciate you reaching out about this. Hope to get more bugs from you in the future :]

dholbert:
> this has [already been changed](https://bugzilla.mozilla.org/show_bug.cgi?id=1347164) in their code

(The Firefox fix hasn't quite been checked in yet, actually, but the patch is written & I expect it should be in Nightly in the next few days.)

patrickkettner:
thanks for the clarification!

PikadudeNo1:
Question for @dholbert: Will Firefox ESR get the fix as well?

dholbert:
That's a call for our release drivers to make, but in my opinion it definitely should, yes. (And I think they'll agree.)

PR #2189, opened 2017-04-13
Add CSS Grid test for old & new syntaxes

KuraFire:
While CSS Grid gets rolled out, it’s useful to have a way to target
IE11 which does not support @supports() but does support old CSS Grid
syntax, as does the new IE14.

### DIFF
--- a/feature-detects/css/cssgrid.js
+++ b/feature-detects/css/cssgrid.js
@@ -2,18 +2,19 @@
 {
   "name": "CSS Grid (old & new)",
   "property": ["cssgrid", "cssgridlegacy"],
+  "authors": ["Faruk Ates"],
   "tags": ["css"],
   "notes": [{
     "name": "The new, standardized CSS Grid",
     "href": "https://www.w3.org/TR/css3-grid-layout/"
   }, {
-    "name": "The _old_ CSS Grid",
+    "name": "The _old_ CSS Grid (legacy)",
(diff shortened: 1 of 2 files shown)


ryanseddon:
Can you add this to [config-all.json](https://github.com/Modernizr/Modernizr/blob/master/lib/config-all.json) too please

ryanseddon:
Normally we have each test separate but I think this makes sense to put it in one since most people will want both.

feature-detects/css/cssgrid.js
```diff
@@ -0,0 +1,19 @@
...
+  "property": ["cssgrid", "cssgridlegacy"],
+  "tags": ["css"],
+  "notes": [{
+    "name": "The new, standardized CSS Grid",
+    "href": "https://www.w3.org/TR/css3-grid-layout/"
+  }, {
+    "name": "The _old_ CSS Grid",
+    "href": "https://www.w3.org/TR/2011/WD-css3-grid-layout-20110407/"
+  }],
+  "warnings": ["`grid-columns` is only in the old syntax, `grid-column` exists in both and so `grid-template-rows` is used for the new syntax."]
```
patrickkettner: this isn't really a warning. Warnings should be about the detect themselves (i.e. 🚨 giamundo test🚨 , or listing possible edge case failures. Since the end user of modernizr won't need this information, but modernizr developers would, could you move this to a straight up JS comment on line 17?

patrickkettner:
lgtm

PR #2218, opened 2017-06-12
Pointer Events: remove hand.js polyfill, expand notes/description

patrickhlauke:
* hand.js is outdated (not modified since 2014, and codeplex is now
obsolete)
* pointer to MDN documentation and PE Level 2
* expand description with regards to browser support (Gecko is currently
experimental)


patrickkettner:
thanks!

PR #2222, opened 2017-06-26
fix appveyor: npm@v4 doesn't work with node.js 0.10 or 0.12

philippsimon:


patrickkettner:
Thanks for the change! would you be able to include updated nodejs_version in the PR as well? current realse and valid LTRs - thanks so much!

philippsimon:
@patrickkettner after several trials I managed to get it working. A different npm version can be set for each nodejs version.

mAiNiNfEcTiOn:
@philippsimon `npm` 4.3.0 works fine with both 0.10 and 0.12 ;-) (Tested on my closed PR #2230)

patrickkettner:
thanks @philipbordallo!

PR #2223, opened 2017-06-26
fix #1663 - Configuration Options not working

### LINKED ISSUE #1663: Configuration Options not working
I'm using the following code to build modernizr:

``` javascript
var fs = require('fs');
var modernizr = require("modernizr");
var modernizrConfig = JSON.parse('' + fs.readFileSync('config.json'));

modernizr.build(modernizrConfig, function (result) {
    fs.writeFileSync('modernizr.js', result);
});
```

And my `config.json` looks like this:

``` json
{
    "classPrefix": "test",
    "enableJSClass": false,
    "enableClasses": false
}
```

However, when I build modernizr, it still sets css classes.
I look into the resulting build file, and I see this:

``` javascript
// Any settings that don't work as separate modules
// can go in here as configuration.
_config: {
    'classPrefix' : 'test',
    'enableClasses' : true,
    'enableJSClass' : true,
    'usePrefixes' : true
},
```

philippsimon:


rejas:
Could you look into why the checks are failing and add a little more documentation to your PR? Thx in advance,

rejas:
Fixed via #2413 Thx for your initial work @philippsimon !

Issue #1409, opened 2014-08-04
Detection for SVG external content support


bitinn:
Basically, figure out a way to test `xlink:href` support of remote uri. Currently missing from all version of IE and older version of android browser (OS 4.1 and down, not confirmed).

I think this will come in handy now that SVG is gradually becoming the go-to choice when working with icon system, replacing font icons. And as far as I know, only inline SVG and `use` element with remote uri will enable the capability to style SVG with CSS.

Current solution such as https://github.com/jonathantneal/svg4everybody is less than optimal as UA sniffing is involved, `background-image` fallback such as https://github.com/filamentgroup/grunticon are nice but prevent developer from styling individual icon, thus multiple version of the same icons are required.

With a better detection, a better polyfill may then be developed (if any exists already?)

reference:
- http://status.modern.ie/svgexternalcontent?term=svg
- https://github.com/filamentgroup/grunticon/issues/139

ArtemGordinsky:
Did anyone succeed in coming up with such a test?

patrickkettner:
not yet - PRs welcome!

bfred-it:
Consider that [svg4everybody is 1.4KB](https://github.com/jonathantneal/svg4everybody/blob/master/dist/svg4everybody.min.js), it's very likely more performant to just just include the file in your main.js file than loading Modernizr, executing the test and loading one more file.

jdalton:
Related to https://github.com/jonathantneal/svg4everybody/issues/71#issuecomment-131894490.

etylsarin:
+1 for proper detection proposal (no UA sniffing)

PR #2227, opened 2017-07-14
Add support for svg external content

Neatoro:


feature-detects/svg/externalresources.js
```diff
@@ -0,0 +1,18 @@
...
+}
+!*/
+/* DOC
+Detects support for External Resources in SVG.
+
+See [this discussion](https://github.com/Modernizr/Modernizr/issues/1409) regarding SVG external content.
+*/
+define(['Modernizr'], function(Modernizr) {
+  Modernizr.addTest('externalresources', function() {
+    return document.implementation.hasFeature('http://www.w3.org/TR/SVG11/feature#ExternalResourcesRequired', '1.1');
```
rejas: Doesnt that violate  https://github.com/Modernizr/Modernizr/issues/1409#issuecomment-51089302 : 
"unfortunately, modernizr can't really use any external uris, ever. all of our tests need to be self contained and not realiant upon a network connection."

patrickkettner:
No, this is something that is built into browsers. If you Google hasImplementation you'll find a bunch about this

rejas:
I see... https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementation/hasFeature says it always returns true in modern browser since its deprecated
That shouldnt be a problem for this issue since modern browsers seem to support the svg feature anyways.
So @Neatoro would you be willing to update your PR agaisnt the latest master sources to get this merged?

patrickkettner:
@rejas I missed the original context of this in my last reply (was going off of the email). 

This should closed, not merged. It is not actually detecting anything (as that function always returns true)

PR #2229, opened 2017-07-20
Add css custom properties feature detection

mAiNiNfEcTiOn:
# What does this PR do:

* _Attempts_ to add a feature detection for CSS Custom Properties (commonly known as _CSS Variables_) to Modernizr.

It is my first PR here, not sure if it's either the proper way or if I am missing something. If that's so, please provide me with the proper guidance and I'll make sure I follow any contribution rules.

### DIFF
--- a/feature-detects/css/customproperties.js
+++ b/feature-detects/css/customproperties.js
@@ -11,17 +11,7 @@
   }]
 }
 !*/
-define(['Modernizr', 'testStyles', 'test/css/supports'], function(Modernizr, testStyles) {
-  var supportsCssVariables = false;
-
-  if (Modernizr.supports) {
-    var defaultStyle = '#modernizr{background: blue}';
-    var testCase = '@supports (--bg-var: red){#modernizr{background: var(--bg-var);}}';
-    testStyles(defaultStyle + testCase, function(elm) {
-      var computedStyle = window.getComputedStyle(elm);
-      supportsCssVariables = computedStyle.getPropertyValue('background') === 'red';
-    });
-  }
-
-  Modernizr.addTest('customproperties', supportsCssVariables);
+define(['Modernizr'], function(Modernizr) {
+  var supportsFn = (window.CSS && window.CSS.supports.bind(window.CSS)) || (window.supportsCSS);
+  Modernizr.addTest('customproperties', !!supportsFn && supportsFn('color', 'var(--primary)'));
 });
--- a/feature-detects/css/customproperties.js
+++ b/feature-detects/css/customproperties.js
@@ -1,0 +1,27 @@
+/*!
+{
+  "name": "CSS Custom Properties",
+  "property": "csscustomproperties",
+  "caniuse": "css-variables",
+  "tags": ["css"],
+  "builderAliases": ["css_customproperties"],
+  "notes": [{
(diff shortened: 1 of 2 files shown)


aFarkas:
There are multiple errors in your test. It seems that you haven't even run your test in any browser.

1. At no point you set the `--bg-var` to `red`. (You only test wether the browser knows the expression)
2. `background` isa  shorthand property for multiple values, not only `backgroundColor`.
3. `backgroundColor` isn't that easy to test because the computed value isn't `red`, it might be `rgb(255, 0, 0)` or `rgba(255,0,0,1)` or something in that range.
4. You are using `testStyles`, which can be used to create very accurate tests, but it adds a big performance penalty. You should avoid `testStyles` with all costs as long as you can. In case of CSS Custom Properties it is easy (you actually already know the answer).

```js
Modernizr.addTest('customproperties', !!window.CSS && !!window.CSS.supports && window.CSS.supports('--bg-var', 'saasas'))
```

mAiNiNfEcTiOn:
@aFarkas Thank you for your feedback :)

Regarding not having tested on browsers, it is in fact true... However it is not clear what is the procedure for testing in this project, if there's a proper command for it or I just have to build the modernizr bundle and use it on a page testing my approach.

Shouldn't the tests be part of the suite as well?

In any case, your example is not in fact true:

```js
!!window.CSS && !!window.CSS.supports && window.CSS.supports('--bg-var', 'saasas')
```
this above condition will return `false` on Safari 10 (which supports CSS Custom Properties):
screen shot 2017-07-21 at 08 46 02

However, a condition like the one below, returns `true`:
```js
!!window.CSS && !!window.CSS.supports && window.CSS.supports('(--bg-var: saasas)')
```

screen shot 2017-07-21 at 08 47 39

mAiNiNfEcTiOn:
Ok, improved the check to only use the `supports()` or `supportsCSS()`.

Also changed the test to be `supports('color', 'var(--primary)')` as - strangely - it works on Safari 10 compared to the `supports('--var', 'value')` test.

Also tested locally with the build. 

Thanks for the feedback.

mAiNiNfEcTiOn:
Well the `appveyor`'s build fails due to incompatibility with `npm` v5. But I've fixed that in another PR ( #2230 ).

mAiNiNfEcTiOn:
@aFarkas not that I'm implying that you don't care about my PR, butt even I forgot i had opened it until today. 

Doesn't seem a big deal to review and there are even duplicates being created.

aFarkas:
@mAiNiNfEcTiOn 
Thx for your effort. I actually just wanted to help and speed up things, but I'm not really an active member so I forgot to check on this.

In general I don't have full rights to merge this in to the project. Maybe @patrickkettner can help to do the final review.

mAiNiNfEcTiOn:
@aFarkas ok. Thanks for your help, then. :+1:

mAiNiNfEcTiOn:
@patrickkettner I've created this PR in July... would you care to take a look at it? If you don't want to merge it, fine by me I can close it. But would be cool to at least get _some_ feedback.

mAiNiNfEcTiOn:
Screw it... I'm tired of waiting for maintainers to take a look at 5 lines of code and a JSDoc block.

PR #2230, opened 2017-07-21
Downgrades npm to 4.3.0 on appveyor.yml to fix the builds on node 0.10 and 0.12

mAiNiNfEcTiOn:
# What does this PR do:

* Downgrades npm to 4.3.0 since newer versions stop supporting node versions such as 0.10 and 0.12.

One question though, does the project still needs to support these old versions of node?


mAiNiNfEcTiOn:
Closing as this is a duplicate of #2222

PR #2244, opened 2017-09-07
Add test for CSS custom properties.

Kruithne:
MDN ref: https://developer.mozilla.org/en-US/docs/Web/CSS/--*
WC3 ref: https://www.w3.org/TR/css-variables/

Apologies if this is already supported somewhere, I checked around but could find no mention of this feature reference as neither custom CSS properties or variables.

### DIFF
--- a/feature-detects/css/custom-properties.js
+++ b/feature-detects/css/custom-properties.js
@@ -1,0 +1,19 @@
+/*!
+{
+  "name": "CSS Custom Properties",
+  "property": "csscustomproperties",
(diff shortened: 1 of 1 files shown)


mAiNiNfEcTiOn:
@Kruithne I also did a PR on that: https://github.com/Modernizr/Modernizr/pull/2229

Not saying that mine is better,  but at least one of them should definitely be merged 😊

rejas:
Feature got finally merged, you can close your PR now. Thanks for your work!

PR #2254, opened 2017-10-10
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that Modernizr gets 36M hits per month on jsDelivr, so I thought you might like this badge.


ryanseddon:
I'm not sure we want to add this since this isn't the way you should be loading Modernizr, we always recommend a custom build for your needs.

PR #2256, opened 2017-10-17
Add Font Display detection

jbmoelker:
Adding detection as [requested by @paulirish](https://twitter.com/paul_irish/status/920073974654033920). Credits go to [@heyam](https://twitter.com/heycam) for the [detection script itself](https://bugzilla.mozilla.org/show_bug.cgi?id=1296373#c4) and [vijayaraghavanramanan](https://github.com/vijayaraghavanramanan) for raising the issue in the first place.

I added a [demo of the `font-display` feature detection](http://plnkr.co/edit/OoxwMS?p=preview) with a Modernizr build containing the new detection.

I was unclear if and how I should [use the `testStyles` helper like done in `css/fontface.js`](https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/fontface.js#L39-L45). I'm happy to make changes if needed.

### DIFF
--- a/feature-detects/css/fontdisplay.js
+++ b/feature-detects/css/fontdisplay.js
@@ -1,0 +1,30 @@
+/*!
+{
+  "name": "Font Display",
+  "property": "fontdisplay",
+  "notes": [{
+    "name": "W3C CSS Fonts Module Level 4",
+    "href": "https://drafts.csswg.org/css-fonts-4/#font-display-desc"
+  },{
+    "name": "`font-display` for the masses",
+    "href": "https://css-tricks.com/font-display-masses/"
+  }]
+}
+!*/
+/* DOC
+Detects support for the `font-display` descriptor, which defines how font files are loaded and displayed by the browser.
+*/
+define(['Modernizr'], function(Modernizr) {
+  Modernizr.addTest('fontdisplay', function() {
+    try {
+      var e = document.createElement('style');
+      e.textContent = '@font-face { font-display: swap; }';
+      document.documentElement.appendChild(e);
+      var isSupported = e.sheet.cssRules[0].cssText.indexOf('font-display') != -1;
+      e.remove();
+      return isSupported;
+    } catch (e) {
+      return false;
+    }
+  });
+});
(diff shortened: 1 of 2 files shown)


jbmoelker:
Correct me if I'm wrong, but looking at the Travis logs the build doesn't appear to fail due to changes made in this PR.

patrickkettner:
@jbmoelker so sorry for taking such a long time to reply.
This is a great idea! We can simplify it a bit, however. You would want to change this to

 `Modernizr.addTest('textshadow', testProp(fontDisplay', 'swap'));`

rejas:
@jbmoelker the travis build errors should be fixed now, could you add the suggestion from @patrickkettner  to your PR so we can move it forward?

rejas:
@patrickkettner tried your approach (fixed a typo I think):

```
define(['Modernizr'], function(Modernizr, testProp) {
  Modernizr.addTest('font-display', testProp('font-display', 'swap'));
});
```

Alas tests fails then:
```
596 passing (4s)
  1 failing

  1) classes everythings ship shape font-display is correctly false in the classes and object:
     Error: expected undefined to not equal undefined
```

Tests don't fail with @jbmoelker approach, si I'd rather merge this PR. Maybe testprops doesnt work because font-display is specified underneath font-face ?

patrickkettner:
It is failing because the first argument to addTest needs to be camelcase

rejas:
Thx for the hint, alas Modernizr doesnt report correctly, for example Chrome and FF report "fontdisplay: false" when I look at the cowtests. Still an issue becuase it isnt a "normal" css property but only used insisde font-face definitions?

patrickkettner:
@rejas sorry for the confusion. 'swap' is only valid within a font-face declaration, but simply checking for font-display support is enough. hence

```js
Modernizr.addTest('fontDisplay', testProp(fontDisplay'));
```

gives the correct result

rejas:
Thx for the initiative to get this started @jbmoelker and thx for the support and help @patrickkettner. Opened a PR for this (#2323) and will merge it.

PR #2269, opened 2017-11-24
Feature detect for text-decoration styling

stefanpl:
Adding a feature detect for text-decoration styling. It is based on text-decoration-color – if this property is enabled in the browser (or isn't), it is very likely that the other text-decoration properties (text-decoration-style, text-decoration-line, text-decoration-skip, text-decoration) follow suit.


rejas:
Thx for your input @stefanpl, merged it with another test and opened a new PR for that: https://github.com/Modernizr/Modernizr/pull/2327

Issue #1196, opened 2014-01-22
Video autoplay error: Cannot call method 'removeChild' of null


robwierzbowski:
Running Modernizr 3.0.0pre in Chrome I get the error:

``` js
elem.parentNode.removeChild(elem); // <- Uncaught TypeError: Cannot call method 'removeChild' of null
```

coming from [line 30 of the video auto play test](https://github.com/Modernizr/Modernizr/blob/ff40a84fdde704a53d6a762767e9a1fbb90f0456/feature-detects/video/autoplay.js#L30).

samccone:
Been plagued by this one for a few months now, able to reproduce intermittently on 

Chrome Version 43.0.2357.130 (64-bit) osx


![screen shot 2015-07-01 at 7 33 04 pm](https://cloud.githubusercontent.com/assets/883126/8468946/08aa88b8-2028-11e5-8d35-b6ca0e504c25.png)

PR #2274, opened 2017-12-22
Ensure elem.parentNode exists before remove its child

simongareste:
Based on feature-detects/video/autoplay.js, ensure that elem is still present.
Linked to issue https://github.com/Modernizr/Modernizr/issues/1196 I believe.

We had all errors regarding this reported only on Chrome 57.0.2987, under Windows 7. I wasn't able to reproduce, as I couldn't figure out how to get this specific version of Chrome :(


patrickkettner:
cheers!

PR #2291, opened 2018-02-12
Remove ancient (and expensive) logic for verifying CSS transform support

bholley:
This code was added in 2009 via [1] to work around a false-positive in WebKit.
Per [2] both Safari and Chrome have supported transforms since 2009 and 2011
respectively, and current global usage of earlier versions of those products
is negligible. Also note that this code originally attempted to limit itself
to WebKit, but the prefixed properties it detects have since been standardized
and implemented in all browsers.

This code is responsible for significant slowdowns (in the tens of tens of
milliseconds) when loading CNN.

[1] https://github.com/Modernizr/Modernizr/issues/15
[2] https://caniuse.com/#search=perspective

### DIFF
--- a/feature-detects/css/transforms3d.js
+++ b/feature-detects/css/transforms3d.js
@@ -9,37 +9,8 @@
   ]
 }
 !*/
-define(['Modernizr', 'testAllProps', 'testStyles', 'docElement', 'test/css/supports'], function(Modernizr, testAllProps, testStyles, docElement) {
+define(['Modernizr', 'testAllProps', 'test/css/supports'], function(Modernizr, testAllProps) {
   Modernizr.addTest('csstransforms3d', function() {
-    var ret = !!testAllProps('perspective', '1px', true);
-    var usePrefix = Modernizr._config.usePrefixes;
-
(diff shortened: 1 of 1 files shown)


ryanseddon:
@patrickkettner do you have any capacity to merge and release this?

patrickkettner:
Just so it is documented somewhere, even though this technically breaks semver, I am going to release this as 3.5.1. Thought being that even though this is a breaking change, and therefor should be a major release, its unlikely to hit anyone who actually would care about this change and folks would be less likely to update.

bholley:
That sounds great, thanks! Let me know when it's ready, and I'll have our WebCompat folks poke CNN about updating their Modernizr (which is currently at 3.3.1).

patrickkettner:
@bholley it was released shortly after that comment as v3.6.0 :)

bholley:
Ah, perfect! I thought you'd released it as 3.6.0 and then changed your mind and wanted to re-release it, but I think it was actually the other way around.

Thanks a ton for getting this over the line - it'll make the web ever-so-slightly better for everyone. :-)

PR #2299, opened 2018-03-12
Passive event listener detection causes error for EdgeHTML 15

mhadam:
This seems to be an issue for Edge because it breaks the DOM standard, but still worth fixing in my opinion.


patrickkettner:
hey @mhadam!
Thanks a bunch for the PR. We are not using ES6 at the moment, however. Would you be wiling to change it back to ES5?

mhadam:
@patrickkettner Ok - new commit should have fixed it!

PR #2302, opened 2018-03-22
Ensure workers exists before terminating them

simongareste:
This should avoid the "undefined is not a function"  error when terminating workers


patrickkettner:
@simongareste where did you see this error? If they are `undefined` they should not pass the current `if` since `undefined` is falsey

rejas:
hi @simongareste is this still a thing? could you provide some more infos as requested by @patrickkettner ?

PR #2294, opened 2018-02-16
Cleanup and Update some dependencies

rejas:
Hi, took the liberty to update the dependencies except jade and eslin. Tthose two would take a little more eyesight to see if somethign breaks: jade got renamed, eslint seems to dislike preset, jsdoc etc in its newest version.

PR #2309, opened 2018-04-29
touchevents: Use mq in touch event detection.

emilio:
testStyles + offsetTop is unnecessarily slow if the browser supports matchMedia.

### DIFF
--- a/feature-detects/mediaquery/hovermq.js
+++ b/feature-detects/mediaquery/hovermq.js
@@ -12,5 +12,5 @@
 Detect support for Hover based media queries
 */
 define(['Modernizr', 'addTest', 'mq'], function(Modernizr, addTest, mq) {
-  Modernizr.addTest('hovermq', mq(('(hover)')));
+  Modernizr.addTest('hovermq', mq('(hover)'));
 });
--- a/feature-detects/touchevents.js
+++ b/feature-detects/touchevents.js
@@ -33,20 +33,16 @@ It's recommended to bind both mouse and touch/pointer events simultaneously – 
 
 This test will also return `true` for Firefox 4 Multitouch support.
 */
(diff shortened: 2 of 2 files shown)


emilio:
The Travis and AppVeyor failures look unrelated (and the Travis + node 6 run is green), but let me know if they're not.

feature-detects/mediaquery/hovermq.js
```diff
@@ -12,5 +12,5 @@
 Detect support for Hover based media queries
 */
 define(['Modernizr', 'addTest', 'mq'], function(Modernizr, addTest, mq) {
-  Modernizr.addTest('hovermq', mq(('(hover)')));
```
patrickkettner: why this change?

emilio: I just saw the extra parentheses while I was going through similar patterns. I can remove it if needed, but I thought a bit of cleanup never hurts. Let me know if you want to remove that commit from this PR and submit it separately.

patrickkettner: no we're good, just making sure there wasn't an issue

rejas:
@emilio @patrickkettner the PR https://github.com/Modernizr/Modernizr/pull/2294 would fix the Travis and AppVeyor failures

bholley:
@emilio what's the status here?

bholley:
Ok, great! I just remeasured, and confirmed that this fix still saves ~15ms on cnn pageload.

@patrickkettner, any chance y'all could do another dot release (similar to #2291) so that we can get this into CNN's pipeline?

rejas:
would love to see a v3.7 out the door too, alas we are still behind with the release notes for 3.6 :-(

PR #2312, opened 2018-05-03
Add detect for the GetBattery API

abrad1212:


feature-detects/battery.js
```diff
@@ -16,5 +16,5 @@
 Detect support for the Battery API, for accessing information about the system's battery charge level.
 */
 define(['Modernizr', 'prefixed'], function(Modernizr, prefixed) {
-  Modernizr.addTest('batteryapi', !!prefixed('battery', navigator), {aliases: ['battery-api']});
+  Modernizr.addTest('batteryapi', !!prefixed('getBattery', navigator), {aliases: ['battery-api']});
```
patrickkettner: as both shipped in browsers, we want to test for both.

abrad1212: @patrickkettner Ok so a new file should be created?

patrickkettner:
sorry, I wasn't clear. Not a new detect, both versions of the API should be detected by the same `battery` detect

abrad1212:
@patrickhlauke Ok. I think I understand, I just don't know how to implement it (I'm fairly new to this library). :P

patrickkettner:
@abrad1212 you would just do both detects, with an `||` in between

```js
!!prefixed('battery', navigator) || !!prefixed('getbattery', navigator)
```

you would also want to change the file and detect name back to just `battery`

feature-detects/battery.js
```diff
@@ -16,5 +16,5 @@
 Detect support for the Battery API, for accessing information about the system's battery charge level.
 */
 define(['Modernizr', 'prefixed'], function(Modernizr, prefixed) {
-  Modernizr.addTest('batteryapi', !!prefixed('battery', navigator), {aliases: ['battery-api']});
+  Modernizr.addTest('batteryapi', !!prefixed('battery', navigator) || !!prefixed('getbattery', navigator), {aliases: ['battery-api']});
```
patrickkettner: sorry, didn't notice before. But this should be `getBattery`, not `getbattery`

abrad1212: Oh lol, I should have known. :P Thanks

PR #2315, opened 2018-05-17
Fix exception in indexedDB check on Safari

UCIS:
We frequently receive error reports like "SecurityError: IDBFactory.open() called in an invalid security context" which originate in the indexedDB check on Safari browsers (iOS and Mac).

It appears that Safari throws an exception on the IDBFactory.open call when running in an iframe on another domain.


steffenweber:
We are using Modernizr because it is bundled with our forum software (XenForo). Among the 112 JavaScript errors that were logged in the past 24 hours, 81 are related to this issue: Calling `indexeddb.open` throws an "InvalidStateError" or an "UnknownError" in different Firefox versions (most of them are Firefox 52, so maybe there is something special about the ESR versions).

The relevant commit in this pull request seems to be https://github.com/Modernizr/Modernizr/pull/2315/commits/4708e859bbcbc8c6cae6f4396c22dcd2682bbadc. I'd appreciate if you could cherry-pick it.

rejas:
Hi @steffenweber made a new PR (https://github.com/Modernizr/Modernizr/pull/2336) with the change. Can you test if that branch works for you better?

steffenweber:
XenForo ships a minified version of Modernizr but I've now manually applied the patch to the minified version (I've replaced `,o=t.open(r)` with `;try{o=t.open(r);}catch(n){i("indexeddb",!0);return;}`).

I'll let you know in a few hours (or tomorrow) whether the has caused the error log entries to go away.

steffenweber:
Unfortunately, it seems like this doesn't help. But according to a Stack Overflow comment, calling event.preventDefault() in the error handler helps. I'm trying that now and will let you know if that solves the problem.

https://stackoverflow.com/a/45672895
https://bugzilla.mozilla.org/show_bug.cgi?id=1331103#c3

rainman1100:
I sent you a Facebook sticker can I tell who it came from and who sent it

steffenweber:
Calling `event.preventDefault()` fixes the problem. I've created a pull request: https://github.com/Modernizr/Modernizr/pull/2337 :smiley: 

@UCIS I'm not sure whether this also fixes your issue. I'm sorry for hijacking your pull request, I initially thought these issues were related but I'm not sure about that anymore.

UCIS:
I did not have much time to update the PR. Thanks @rejas for updating the fix. @steffenweber I don't have a Mac system to test with so I can't check whether your PR would fix this problem, but I think it won't, since this PR fixes a synchronous SecurityError.

PR #2318, opened 2018-05-28
Remove some no-op code in webgl.js

### LINKED ISSUE #2300: probablySupportsContext is removed from HTML and not implemented
It seems you can remove this from https://github.com/Modernizr/Modernizr/blob/master/feature-detects/webgl.js as it's not being run.

Source: https://github.com/whatwg/html/issues/3572.

annevk:
Fixes #2300.

### DIFF
--- a/feature-detects/webgl.js
+++ b/feature-detects/webgl.js
@@ -7,13 +7,8 @@
   "polyfills": ["jebgl", "cwebgl", "iewebgl"]
 }
 !*/
-define(['Modernizr', 'createElement'], function(Modernizr, createElement) {
+define(['Modernizr'], function(Modernizr) {
   Modernizr.addTest('webgl', function() {
-    var canvas = createElement('canvas');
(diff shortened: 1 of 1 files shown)


feature-detects/webgl.js
```diff
@@ -7,13 +7,8 @@
   "polyfills": ["jebgl", "cwebgl", "iewebgl"]
 }
 !*/
-define(['Modernizr', 'createElement'], function(Modernizr, createElement) {
+define(['Modernizr'], function(Modernizr) {
   Modernizr.addTest('webgl', function() {
-    var canvas = createElement('canvas');
-    var supports = 'probablySupportsContext' in canvas ? 'probablySupportsContext' :  'supportsContext';
```
rejas: is supportsConext also removed or why did you delete it too?

annevk: That was the original name of the feature, so yes, it's also removed: https://github.com/whatwg/html/commit/2542a12cb25ee93534cbed1f31b5e1bc05fcdd0e.

PR #2329, opened 2018-06-09
Replace == with ===

rejas:
webstorm noticed a few instances with == where === would be more appropiate. did the same with != -> !==


patrickkettner:
you shouldn't be touching library code like sinon, is there a reason for it?

rejas:
no, that was just an oversight. will reverse those parts...

patrickkettner:
LGTM assuming it doesnt break travis and appveyor

PR #2333, opened 2018-06-15
Js doc cleanup

rejas:
Lots of http: --> https: conversions (still some left to do though or replace with correct notes)
Cleaned up some tags

### DIFF
--- a/feature-detects/audio/webaudio.js
+++ b/feature-detects/audio/webaudio.js
@@ -8,7 +8,7 @@
   "builderAliases": ["audio_webaudio_api"],
   "authors": ["Addy Osmani"],
   "notes": [{
-    "name": "W3 Specification",
+    "name": "W3 Spec",
     "href": "https://dvcs.w3.org/hg/audio/raw-file/tip/webaudio/specification.html"
   }]
 }
--- a/feature-detects/css/flexbox.js
+++ b/feature-detects/css/flexbox.js
@@ -6,7 +6,7 @@
   "tags": ["css"],
   "notes": [{
     "name": "The _new_ flexbox",
-    "href": "https://drafts.csswg.org/css-flexbox-1/"
+    "href": "https://www.w3.org/TR/css-flexbox-1/"
   }],
   "warnings": [
     "A `true` result for this detect does not imply that the `flex-wrap` property is supported; see the `flexwrap` detect."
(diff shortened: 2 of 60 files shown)


feature-detects/audio/webaudio.js
```diff
@@ -8,7 +8,7 @@
   "builderAliases": ["audio_webaudio_api"],
   "authors": ["Addy Osmani"],
   "notes": [{
-    "name": "W3 Specification",
+    "name": "W3 Spec",
```
patrickkettner: any particular reason for this?

feature-detects/css/flexbox.js
```diff
@@ -6,7 +6,7 @@
   "tags": ["css"],
   "notes": [{
     "name": "The _new_ flexbox",
-    "href": "http://dev.w3.org/csswg/css3-flexbox"
+    "href": "https://drafts.csswg.org/css-flexbox-1/"
```
patrickkettner: @gregwhitworth while we are updating, would you consider the css drafts release to be the most correct one to link to in modernizr's docs?

feature-detects/elem/picture.js
```diff
@@ -6,7 +6,7 @@
   "authors": ["Scott Jehl", "Mat Marquis"],
   "notes": [{
     "name": "Specification",
-    "href": "http://picture.responsiveimages.org"
+    "href": "http://w3c.github.io/html/semantics-embedded-content.html#the-picture-element"
```
patrickkettner: 👎 on this change. The w3c html spec is dead, so we should not be linking to it for documentation. If you are against linking to the RIWG site as it is a redirect, then please update it to the site it is redirecting to

feature-detects/es5/strictmode.js
```diff
@@ -2,10 +2,10 @@
 {
   "name": "ES5 Strict Mode",
   "property": "strictmode",
-  "caniuse": "sctrict-mode",
```
patrickkettner: lol

feature-detects/audio/webaudio.js
rejas: was the only occasion where "Specification" was used, all other occurences were "W3 Spec" Just my little tidiness-spleen

feature-detects/elem/picture.js
rejas: will change to the redirected page...

feature-detects/es5/strictmode.js
rejas: by the way: I cant find a usage of the caniuse property, is that a TODO or where is it used?

PR #2334, opened 2018-06-15
Fixed some define tags

rejas:


feature-detects/htmlimports.js
```diff
@@ -19,7 +19,6 @@
 /* DOC
 Detects support for HTML import, a feature that is used for loading in Web Components.
  */
-
-define(['addTest', 'createElement'], function(addTest, createElement) {
-  addTest('htmlimports', 'import' in createElement('link'));
+define(['Modernizr', 'addTest', 'createElement'], function(Modernizr, addTest, createElement) {
+  Modernizr.addTest('htmlimports', 'import' in createElement('link'));
```
patrickkettner: why this change?

rejas: I found it strange that this was the only place where addTest was called this way, all other tests use Modernizr.addTest. Was there a reason for this? If so, shouldnt it be documented?

PR #2327, opened 2018-06-07
More test around text decoration styling

rejas:
Merged PR https://github.com/Modernizr/Modernizr/pull/2269 with existing test for skipInk. 

Weirdness: textDecorationColor doesnt show up on the CowTest page but classes are added....

PR #2337, opened 2018-06-23
IndexedDB: Fix InvalidStateError and UnknownError in Firefox

steffenweber:
In Firefox, the method "IDBFactory.open()" can raise an asynchronous error that can only be supressed by calling "event.preventDefault()". See https://github.com/Modernizr/Modernizr/pull/2315.

https://bugzilla.mozilla.org/show_bug.cgi?id=1331103#c3

https://stackoverflow.com/q/13972385


rejas:
@patrickkettner would love to bring v3.7 to a finish, so this and https://github.com/Modernizr/Modernizr/pull/2327 are still open IMO. Any comments?

steffenweber:
Thanks!

Issue #2261, opened 2017-10-31
hyphens test uses invalid CSS property


lolmaus:
On these two lines `text-justification:newspaper` is used:

https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/hyphens.js#L61
https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/hyphens.js#L67

But there's no such property.

Should probably be [text-justify](https://developer.mozilla.org/en-US/docs/Web/CSS/text-justify) but I dunno if such change is safe for the test.

rejas:
Changed it in PR https://github.com/Modernizr/Modernizr/pull/2339 doesnt seem to affect the tests

PR #2339, opened 2018-07-01
Replace text-justification -> text-justify

rejas:
Fixes https://github.com/Modernizr/Modernizr/issues/2261 
text-justification is an unknown css property, should be text-justify. ('newspaper' value is only supported in IE (https://msdn.microsoft.com/en-us/ie/ms531172(v=vs.94)))


patrickkettner:
just for history, can you explain what the change is n a meaningful way here in the issues then merge away afterwards

PR #2342, opened 2018-07-05
Fix testPropsAll testAllProps confusion

### LINKED ISSUE #1645: testAllProps set twice
`ModernizrProto.testAllProps = testPropsAll;`, and then a few lines later, `ModernizrProto.testAllProps = testAllProps;` (or in a minified version, I have `…,w.testAllProps=v,w.testAllProps=y,…`). Sorry if I've misunderstood something, but it doesn't seem right - especially as the function has a different signature each time.

rejas:
Fix #1645

### DIFF
--- a/src/testAllProps.js
+++ b/src/testAllProps.js
@@ -36,6 +36,7 @@ define(['ModernizrProto', 'testPropsAll'], function(ModernizrProto, testPropsAll
    * testAllProps('shapeOutside', 'content-box', true);
    * ```
    */
+
   function testAllProps(prop, value, skipValueTest) {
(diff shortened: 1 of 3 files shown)


rejas:
Any comments on this @patrickkettner or was the testAllProps <-> testPropsAll mixup intentionally done ?

patrickkettner:
Appreciate the PR, but this is intentional. `testPropsAll` is `testProps`, but for all.

dracos:
I hadn't seen the PR before, but the issue I raised in my bug report was not really about the naming of the function, but the fact there are two `testAllProps` , with different function signatures.

PR #2344, opened 2018-07-06
Update require js

rejas:
Uglify2 is now the default minifier and therefore simply called uglify.
(see http://jrburke.com/2016/03/17/requirejs-2.2-alameda-1.0-released/)


rejas:
Comes along with a little space saving too:
94889 bytes vs 94693 bytes in a minified config-all build

Alexkho82:
#2344

PR #2345, opened 2018-07-09
Fixed csshypens detection on FireFox and Safari.

anenviousguest:
# `csshyphens` detection fix for FF and Safari
This is a split version of #1428 as requested by @rejas.

- ~~There was seemingly a typo in CSS property name (`text-justification` instead of correct [`text-justify`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-justify))~~ Addressed in #2339.
- The hyphenation doesn't take effect unless the `lang` attribute is explicitly set, for this reason it was erroneously detected as not supported on FF and Safari, whereas it **is** actually supported.

### DIFF
--- a/feature-detects/css/hyphens.js
+++ b/feature-detects/css/hyphens.js
@@ -52,19 +52,22 @@ define(['Modernizr', 'prefixes', 'createElement', 'testAllProps', 'addTest'], fu
           var result = false;
           var firstChild = document.body.firstElementChild || document.body.firstChild;
 
+          /* Hyphenation is only applied when language is explicitly set and when respective dictionary
+           * is available. See https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens for details. */
+          div.lang = 'en';
           div.appendChild(span);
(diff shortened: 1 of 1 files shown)


feature-detects/css/hyphens.js
```diff
@@ -52,19 +52,22 @@ define(['Modernizr', 'prefixes', 'createElement', 'testAllProps', 'addTest'], fu
...
 
           document.body.insertBefore(div, firstChild);
 
           /* get size of unhyphenated text */
-          divStyle.cssText = 'position:absolute;top:0;left:0;width:5em;text-align:justify;text-justification:newspaper;';
+          divStyle.cssText = 'position:absolute;top:0;left:0;width:5em;text-align:justify;text-justify:newspaper;';
```
rejas: that part is already done in https://github.com/Modernizr/Modernizr/pull/2339/files

feature-detects/css/hyphens.js
```diff
@@ -52,19 +52,22 @@ define(['Modernizr', 'prefixes', 'createElement', 'testAllProps', 'addTest'], fu
...
           spanHeight = span.offsetHeight;
           spanWidth = span.offsetWidth;
 
           /* compare size with hyphenated text */
           divStyle.cssText = 'position:absolute;top:0;left:0;width:5em;text-align:justify;' +
-            'text-justification:newspaper;' +
+            'text-justify:newspaper;' +
```
rejas: this part is also done in https://github.com/Modernizr/Modernizr/pull/2339/files

feature-detects/css/hyphens.js
```diff
@@ -52,19 +52,22 @@ define(['Modernizr', 'prefixes', 'createElement', 'testAllProps', 'addTest'], fu
           var result = false;
           var firstChild = document.body.firstElementChild || document.body.firstChild;
 
+          /* Hyphenation is only applied when language is explicitly set and when respective dictionary
+           * is available. See https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens for details. */
+          div.lang = 'en';
```
rejas: good find. just for documentation purposes: does it fix any special issue (in this github repo or in your work)?

anenviousguest: I think I was puzzled why the test page reports unsupported hyphenation on Safari and FF while they both actually support it.

anenviousguest:
Hi @rejas , adjusted as per your comments.

PR #2347, opened 2018-07-10
Update jquery

rejas:
Trying out to update jquery to v3.3, not sure yet why tests fail

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -243,13 +243,10 @@
       "dev": true
     },
     "async": {
-      "version": "2.6.2",
-      "resolved": "https://registry.npmjs.org/async/-/async-2.6.2.tgz",
-      "integrity": "sha512-H1qVYh1MYhEEFLsP97cVKqCGo7KfCyTt6uEWqsTBr9SO84oK9Uwbyd/yCW+6rKJLHksBNUVWZDAjfS+Ccx0Bbg==",
-      "dev": true,
-      "requires": {
-        "lodash": "^4.17.11"
-      }
+      "version": "1.5.2",
+      "resolved": "https://registry.npmjs.org/async/-/async-1.5.2.tgz",
+      "integrity": "sha1-7GphrlZIDAw8skHJVhjiCJL5Zyo=",
+      "dev": true
     },
     "async-limiter": {
       "version": "1.0.0",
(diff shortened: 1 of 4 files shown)


patrickkettner:
Jquery is only used in the test suite, and does not support the browsers that the test suite runs against

rejas:
But why does it break with v3.3 then and not already with a 3.x release? https://blog.jquery.com/2018/01/19/jquery-3-3-0-a-fragrant-bouquet-of-deprecations-and-is-that-a-new-feature/ looks innocent enough for a minor version jump

patrickkettner:
Dunno. I can check it out when I get back stateside

rejas:
I hope I didnt disturbe you on your summer vacation :-) When you're back: which browsers do need to be supported by the test suite? expect.js seems outdated, maybe even the grunt-test pipeline (but that's just some rambling, not sure if its worth...)

patrickkettner:
not at all! I was at a conference for work, just didn't have a lot of free time for OS stuff during it

not sure what you mean though. expect is old, but it works in all browsers and baring an actual bug or gigantic perf issue, I don't see why we would change for changes sake. More or less same comment about grunt.

As for browsers, its a bit tricky. Folks that are only concerned about modern browsers don't need Modernizr for a vast majority of cases. point being, it is really the most useful for older browsers. What are your thoughts?

rejas:
So this jquery issue aside (hope you can take alook at it) there are issues with grunt and mocha. but I cannot get those updated with more errors popping up (and some grunt plugins not being repsonsive or maintained anymore) thats why I was rambling about if a change of libraries would be a thing....

For example chai was a try to get some progress in this regard because due to the syntax it was easy to switch to (minus 4 tests that still failed for whatever reason).

My intermediate goal is to get the project running with latest version of its dependencies (due to npm audit issues `found 19 vulnerabilities (6 low, 9 moderate, 3 high, 1 critical)`) with node 10. If its done grunt or gulp, chai or expect, doesnt matter in the end to me, I jsut want to have a stable dev environments to attract other developers. Having a "dont use node10 or this or that os" policy doesnt help IMHO.

rejas:
Opened an issue for this: https://github.com/Modernizr/Modernizr/issues/2406

PR #2352, opened 2018-07-26
Replace defunct caniuse jsonp.php?callback

rejas:
I noticed that the integration tests try to use that callback function to get caniuse data and test the browser against it. Seems though that this API is gone for sometime already, openened an [issue](https://github.com/Fyrd/caniuse/issues/4386) asking about it.

In the meantime I took the liberty to get the data directly from their github. 

Noticed though that running the tests on Firefox they fail for quite some cases. Turns out the code thinks it has to test against the Firefox 3.6 (which obviously isnt correct). 

This line:
https://github.com/Modernizr/Modernizr/blob/c2906e9eaec0fa7762eb99b99199b1528daa499a/test/browser/integration/caniuse.js#L315 
 seems doesnt return the keys (via _keys()) in the original order so "3.6" is the last one in the retuned value and therefor gets taken in the _findLast call.

ANy advice on how to get the keys in the original order @patrickkettner ?


rejas:
Fixed the order of the version keys problem.

rejas:
Fairly done for now, would love to hear feedback from @patrickkettner 

Running our tests against the caniuse-db shows that only a few are wrong (unclear on which side):
- IE11: Caniuse result for svg-img matches Modernizr's result for svgasimg 
- Edge17: Caniuse reported partial support for indexeddb
- Firefox61: Caniuse result for ambient-light matches Modernizr's result for ambientlight 

Will open a seperate ticket for those.

PR #2353, opened 2018-07-27
Autogenerate info for license file

### LINKED ISSUE #2289: Autogenerate info for license file
Suggestion from https://github.com/Modernizr/Modernizr/pull/2279#issuecomment-361659536 

Instead of manually adding the version info to the license file it should be somehow generated automatically.

niksy:
Closes #2289.


rejas:
lgtm! but then @patrickkettner should say how envisioned when and how the version generation should be run.

PR #2357, opened 2018-07-31
Use package.json `files` instead of .npmignore

niksy:
It’s easier to explicitly define *what should be published* instead of excluding it.


rejas:
lgtm!

patrickkettner:
thanks!

PR #2359, opened 2018-08-02
Move custom ESLint rule to `no-restricted-syntax`

niksy:
`rulePaths` from ESLint Node API is only available when directly used, but if you have linter active in code editor, Modernizr custom rule doesn’t work since it’s not discoverable. `.eslintrc` currently doesn’t have support for rule options ([open issue](https://github.com/eslint/eslint/issues/2715)), but we can use `no-restriced-syntax` with AST selectors to have that rule applied universally.


rejas:
lgtm!

PR #2363, opened 2018-08-10
option to specify global var name to attach modernizr to #2362

mreinstein:


rejas:
Nice work. Could you also add some kind of test for this?

mreinstein:
Yeah good catch, sorry about that.

Issue #210, opened 2011-03-03
onInput event detection support


thatmarvin:
It'll be nice if Modernizr can detect support for the <code>input</code> event [1]. At the moment, at least Gecko and WebKit supports it.

[1] http://www.whatwg.org/specs/web-apps/current-work/multipage/common-input-element-attributes.html#common-event-behaviors

johan:
These days, there's about three levels of input events support: I guess the initial `oninput` this probably is about (input events that don't have an `inputType` property) that Firefox supports might be called InputEvents 0, and then there's the latter type specified by the W3C specs [InputEvents Level 1](https://www.w3.org/TR/input-events-1/) and  [InputEvents Level 2](https://www.w3.org/TR/input-events-2/) (that also add the `beforeinput` event, which I don't know of any intuitive more direct way to detect) that I'd suggest this hack to detect:

```js
const input_events_level1_or_better = 'inputType' in (new InputEvent('input'));
```

PR #2368, opened 2018-08-25
Issue 2338 - Cleanup JSdoc

Nicholas-Eng:


feature-detects/xdomainrequest.js
```diff
@@ -7,11 +7,7 @@
   "notes": [{
     "name": "MDN documentation",
     "href": "https://developer.mozilla.org/en-US/docs/Web/API/XDomainRequest"
-  },
-  {
-    "name": "MSDN documentation",
-    "href": "https://msdn.microsoft.com/library/ie/cc288060.aspx/"
-  }]
+  }
```
rejas: missing a "]" here which breaks the tests

feature-detects/elem/ruby.js
```diff
@@ -8,7 +8,7 @@
   "authors": ["Cătălin Mariș"],
   "notes": [{
     "name": "WHATWG Specification",
```
rejas: total nitpicking: since you unified the w3c naming, maybe you can do it with WHATWG / WhatWG Spec /Specification too?

feature-detects/elem/template.js
```diff
@@ -7,8 +7,8 @@
     "name": "HTML5Rocks Article",
     "href": "https://www.html5rocks.com/en/tutorials/webcomponents/template/"
   },{
-    "name": "W3 Spec",
-    "href": "https://dvcs.w3.org/hg/webcomponents/raw-file/tip/spec/templates/index.html"
+    "name": "W3C Spec",
+    "href": "https://web.archive.org/web/20170515055729/http://www.w3.org/TR/html5/scripting-1.html"
```
rejas: just for info: when retrieving from web-archive-org did you take the last available useful date?

feature-detects/event/oninput.js
```diff
@@ -7,10 +7,10 @@
     "href": "https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers.oninput"
   },{
     "name": "WHATWG spec",
-    "href": "https://html.spec.whatwg.org/multipage/forms.html#common-input-element-attributes"
+    "href": "https://html.spec.whatwg.org/multipage/input.html#common-input-element-attributes"
   },{
-    "name": "Detecting onInput support",
-    "href": "http://danielfriesen.name/blog/2010/02/16/html5-browser-maze-oninput-support"
+    "name": "Undetectable",
+    "href": "https://github.com/Modernizr/Modernizr/wiki/Undetectables"
```
rejas: maybe point to the issue mentioned in the wiki directly? https://github.com/Modernizr/Modernizr/issues/210

feature-detects/contextmenu.js
```diff
@@ -8,7 +8,7 @@
     "href": "https://www.w3.org/TR/html5/interactive-elements.html#context-menus"
   },{
     "name": "thewebrocks.com Demo",
-    "href": "https://thewebrocks.com/demos/context-menu/"
+    "href": "http://thewebrocks.com/demos/context-menu/"
```
rejas: the only place where we have to go back to http? darn....

Nicholas-Eng: Yes, for some reason http works but not https for this link.

feature-detects/event/oninput.js
Nicholas-Eng: Good point!

feature-detects/elem/template.js
Nicholas-Eng: I tried for most of them, but I I'm gonna double check.

feature-detects/elem/ruby.js
Nicholas-Eng: I agree with this comment. I will try to do this for other links as well.

feature-detects/xdomainrequest.js
Nicholas-Eng: Sorry, that was my mistake!

feature-detects/elem/template.js
rejas: I think it isnt necessary to double check. That would take too much time for little gain.

rejas:
Thx :-)

PR #2379, opened 2018-10-15
PR: Create module_type (Still Editing)

frank-dspeed:
Add support for es6 module support detection

PR #2380, opened 2018-10-15
PR: Create dynamic_import.js (Still Editing)

frank-dspeed:
Add dynamic imports check

### DIFF
--- a/feature-detects/es6/dynamic_import.js
+++ b/feature-detects/es6/dynamic_import.js
@@ -1,0 +1,27 @@
+/*!
+{
+  "name": "ES6 DynamicImport",
+  "property": "import",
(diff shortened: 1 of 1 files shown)


rejas:
thx for the pr @frank-dspeed could you fix the test errors?

rejas:
One more thing: could you add your tests (also https://github.com/Modernizr/Modernizr/pull/2379)  to lib/config-all.json?

rejas:
hi  @frank-dspeed still editing? :-)

frank-dspeed:
@rejas at present i have no idea what is wrong with that meta i am not a heavy modernizer user there was a time as i tought it would be a good idea to use it but we can stop that PR some one else should take over

PR #2381, opened 2018-10-17
Adding feature detection for navigator.connection.effectiveType

### LINKED ISSUE #2378: Adding navigator.connection.effectiveType
Saw navigator.connection feature detection but not seeing this in the docs, on the site, or in the issues.

Basically proposing that detecting the effectiveType would be helpful for determining a more specific kind of connection (slow-2g, 2g, 3g, 4g).

More here: [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/effectiveType) and [here](https://dev.to/addyosmani/adaptive-serving-using-javascript-and-the-network-information-api-331p)

Apologies if this is completely unhelpful... just an idea. Thanks!

escodel:
Closes #2378 

Required a workaround to detect navigator.connection or set it to a value if not present. Detects and returns true if value is not zero. Otherwise, returns false.

- Chrome Version 69.0.3497.100 (result: true)
- Firefox Version 62.0.2 (result: false)
- Edge Version 17.17134 (result: false)

### DIFF
--- a/feature-detects/network/connection-effectivetype.js
+++ b/feature-detects/network/connection-effectivetype.js
@@ -1,0 +1,27 @@
+/*!
+{
+  "name": "Connection Effective Type",
+  "notes": [{
+    "name": "MDN documentation",
+    "href": "https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/effectiveType"
+  }],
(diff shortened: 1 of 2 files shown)


rejas:
Thx :-)

PR #2387, opened 2018-10-20
Update CSS custom properties check to be more robust

### LINKED ISSUE #2383: CSS custom properties check returns false (incorrectly) in Edge
The [custom properties check](https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/customproperties.js) will fail in Edge (tested in version  44.17763.1.0).

The existing code does a check in the form of:
`supportsFn('--f:0')`
This returns `true` correctly in other browsers I've tested that support CSS custom properties, however Edge will return false.
When providing the check in the property/value format, however, Edge correctly returns `true`:
` supportsFn('--f', 0)`
(However Safari [tested in version 12.0] will return true on the first test and false for the second.)

A simple fix would be to check if either of these are true in the custom property support check:
`supportsFn('--f:0') || supportsFn('--f', 0)`

Happy to submit a PR for this if it sounds like the way forward!

nali:
Allows for different browser implementations of the CSS.supports method for custom properties by checking for either a condition, or a property and value set.

Fixes #2383.

### DIFF
--- a/feature-detects/css/customproperties.js
+++ b/feature-detects/css/customproperties.js
@@ -16,5 +16,5 @@
 !*/
 define(['Modernizr'], function(Modernizr) {
   var supportsFn = (window.CSS && window.CSS.supports.bind(window.CSS)) || (window.supportsCSS);
-  Modernizr.addTest('customproperties', !!supportsFn && supportsFn('--f:0'));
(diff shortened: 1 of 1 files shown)


rejas:
Thx :-)

PR #2393, opened 2018-11-07
Adding PublicKeyCredential feature detect for webauthn

### LINKED ISSUE #2391: Detecting PublicKeyCredential feature
Is it possible to use Modernizr to detect the PublicKeyCredential feature (https://developers.google.com/web/updates/2018/03/webauthn-credential-management#the_solution)? This is a new cryptographic web authentication method being standardized by the W3C (https://www.w3.org/TR/webauthn/) and it would be great to use Modernizr to help determine if the browser is capable of supporting the standard.

escodel:
Resolves #2391 

Feature detect for PublicKeyCredential in the Web Authentication API (webauthn). Request/issue was for PublicKeyCredential specifically, but I think this could be expanded to detect more credential types in webauthn as well its parent `Credential`. 

Would this quick Boolean check be adequate? I welcome any feedback or input. Thanks!

### DIFF
--- a/feature-detects/webauthn/publickeycredential.js
+++ b/feature-detects/webauthn/publickeycredential.js
@@ -1,0 +1,30 @@
+/*!
+{
+  "name": "PublicKeyCredential",
+  "notes": [
+    {
+      "name": "MDN Documentation",
+      "href": "https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential"
+    },
+    {
+      "name": "Google Developers solution",
+      "href": "https://developers.google.com/web/updates/2018/03/webauthn-credential-management#the_solution"
+    }
+  ],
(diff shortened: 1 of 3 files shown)


feature-detects/webauthn/publickeycredential.js
```diff
@@ -0,0 +1,30 @@
...
+      "href": "https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential"
+    },
+    {
+      "name": "Google Developers solution",
+      "href": "https://developers.google.com/web/updates/2018/03/webauthn-credential-management#the_solution"
+    }
+  ],
+  "property": "publicKeyCredential",
+  "tags": ["webauthn", "web authentication"]
+}
```
rejas: you can add yourself as an author here :-)

PR #2403, opened 2018-12-28
using the original detection method from to fix Edge (#2310)

rikvanmechelen:
- Fixing detection for MS Edge by using the original proposed method.


rikvanmechelen:
As mentioned [here](https://github.com/Modernizr/Modernizr/pull/2310#discussion_r244363560)
the detection did not work correctly in Edge.
Putting the original code of #2310 back in.

rikvanmechelen:
@rejas I seemed to have missed that pr.

I think that implementation is better and thus supersedes my changes. Closing the pr now.

PR #2420, opened 2019-02-13
Emoji use backingStorePixelRatio instead of devicePixelRatio

### LINKED ISSUE #2419: Emoji - Chrome 72, retina screen: false negative
When I run the emoji detect code on a retina screen (window.devicePixelRatio === 2) on Chrome 72, I get a false negative for Emoji support. This is the relevant code:

```javascript
var pixelRatio = window.devicePixelRatio || 1;
var offset = 12 * pixelRatio;
var node = window.document.createElement('canvas');
var ctx = node.getContext('2d');
ctx.fillStyle = '#f00';
ctx.textBaseline = 'top';
ctx.font = '32px Arial';
ctx.fillText('\ud83d\udc28', 0, 0); // U+1F428 KOALA

var support = ctx.getImageData(offset, offset, 1, 1).data[0] !== 0;
console.log('support?', support);
```

from the most recent [emoji.js file](https://github.com/Modernizr/Modernizr/blob/5e3f359bfc9aa511543ece60bd8a6ea8aa7defd3/feature-detects/emoji.js) (appears to be v3.6.0?)

Here are some results I got testing on Mac OSX 10.13.6:

Browser | devicePixelRatio | emoji support?
--------|------------------|----------------
Chrome 72.0 | 1 | YES
Chrome 72.0 | 2 | NO
Chrome 49.0 | 1 | YES
Chrome 49.0 | 2 | YES
Firefox 65.0 | 1 | YES
Firefox 65.0 | 2 | YES

Based on my understanding of how the canvas 2d context works, I think there’s a bug in the Modernizr code. The devicePixelRatio has no guarantee on the size of a canvas. It looks like some older versions of Safari used to have a backing store pixel ratio and on retina screens the backing store pixel ratio would larger to match the devicePixelRatio, but backing pixel ratio has since been deprecated ([ref](https://stackoverflow.com/questions/24332639/why-context2d-backingstorepixelratio-deprecated)).

This leads me to believe that the code in question was mistakenly treating devicePixelRatio as backingStorePixelRatio in the above code where it determines the offset.

I ran a test to see what I got back for `ctx.getImageData(offset, offset, 1, 1).data` depending on the devicePixelRatio for each browser. Per browser, I got different pixel values between the two devicePixelRatios. When I hardcoded "offset" to `12`, per browser, I got the same exact pixel values between the two devicePixelRatios.

One solution would be to rewrite this code:

```javascript
var pixelRatio = window.devicePixelRatio || 1;
var offset = 12 * pixelRatio;
```

to this:

```javascript
var offset = 12;
```

OR: if we want to support the deprecated backing store pixel ratio, we could do:

```javascript
var node = window.document.createElement('canvas');
var ctx = node.getContext('2d');
var backingStoreRatio =
  ctx.webkitBackingStorePixelRatio ||
  ctx.mozBackingStorePixelRatio ||
  ctx.msBackingStorePixelRatio ||
  ctx.oBackingStorePixelRatio ||
  ctx.backingStorePixelRatio ||
  1;
var offset = 12 * backingStoreRatio;
ctx.fillStyle = '#f00';
ctx.textBaseline = 'top';
ctx.font = '32px Arial';
ctx.fillText('\ud83d\udc28', 0, 0); // U+1F428 KOALA

var support = ctx.getImageData(offset, offset, 1, 1).data[0] !== 0;
```

mrcoles:
While backingStorePixelRatio is deprecated, devicePixelRatio is being used incorrectly in the existing code as the backing store pixel ratio. This code updates the emoji detect to reliably request the same pixel of the canvas regardless of devicePixelRatio (and also supports the rare browser—older Safari—that uses backingStorePixelRatio).

Fixes #2419

### DIFF
--- a/feature-detects/emoji.js
+++ b/feature-detects/emoji.js
@@ -12,10 +12,16 @@ define(['Modernizr', 'createElement', 'test/canvastext'], function(Modernizr, cr
     if (!Modernizr.canvastext) {
       return false;
     }
-    var pixelRatio = window.devicePixelRatio || 1;
-    var offset = 12 * pixelRatio;
     var node = createElement('canvas');
     var ctx = node.getContext('2d');
+    var backingStoreRatio =
+      ctx.webkitBackingStorePixelRatio ||
+      ctx.mozBackingStorePixelRatio ||
+      ctx.msBackingStorePixelRatio ||
+      ctx.oBackingStorePixelRatio ||
+      ctx.backingStorePixelRatio ||
+      1;
+    var offset = 12 * backingStoreRatio;
     ctx.fillStyle = '#f00';
     ctx.textBaseline = 'top';
     ctx.font = '32px Arial';


rejas:
hi @mrcoles sorry for the long silence. could you rebase your PR against the latest master so that we can see if it runs the fixed tests without an error?

mrcoles:
Hi @rejas I just rebased against upstream/master

PR #2426, opened 2019-03-17
Tweak testProps.js documentation to avoid misreading

foolip:
The lack of punctuation/capitalization made "create our own elements to
use inside of an SVG element" a plausible reading at a first pass, but
that's not the intended meaning.


rejas:
Very true :-) Thx for the PR.

Issue #2431, opened 2019-03-21
Desktop chrome touchevent tests returns "true" always.


koskinpark:
Hello there.
I know about https://github.com/Modernizr/Modernizr/pull/2409
But question is "why"? Of course now it works as expected in terms of "description" of this test. But usage of this test was to detect real touch screens, not browser feature. I don't against last fixes for touchevents, but i'd say we need to have both possibilities to check any case:
1. Does this browser support touch feature? // btw, please tell me for what exactly it is useful to use this test?
2. Does current device have the real touch screen?

roblarsen:
See #869 

Also check out [Getting touchy](https://patrickhlauke.github.io/getting-touchy-presentation/#Cover)

gregmatys:
it works incorrectly from 3.7.1 version where `window.TouchEvent` is added

rejas:
Closing this since we deprecated the test with v3.8.0

Issue #2430, opened 2019-03-20
Modernizr 3.7.1 sets .touchevents on Desktop-Chrome


BenRichter:
macOS High Sierra (v 10.13.6)
Chrome Version 72.0.3626.121 (64-bit)
Modernizr v3.7.1

After upgrading modernizr, the class .touchevents (instead of .no-touchevents) was set in Chrome. 
- Firefox and Safari are working fine. 
- Didn't zoom or else
- Incognito-Mode without Plugins has same effect
- Same in colleagues Vivaldi-Browser (chromium)

rejas:
See #2428: The test detects if the browser is capable of detecting touches, not if the device is a touch device.

BenRichter:
My bad. I guess there are hundreds of sites using this feature wrong.
In our case we're initiating https://github.com/ftlabs/fastclick to remove click delays on touch devices, if touchevents are enabled. But this will fire two events now... :/

rejas:
I have to admit I find this confusing too. Maybe it is time to add another test for detecting if the device is touch-capable. https://ctrlq.org/code/19616-detect-touch-screen-javascript seems promising for detecting such devices. What do you think?

niksy:
Just to chime in - I think feature test should tell if device is actually touch device; I think people expect that from this feature test.

rejas:
If they expect that they are clearly not reading the documentation ;-) 

Yes, this test also clearly needs a cleanup / refinement. But I dont think that changing its scope now would help sicne we dont know how many people use it wrongly / correctly. There for adding another tests (and updating the docs to point to that accordingly) would be the better path IMO

SlexAxton:
This is one of the oldest and most recurring requests for modernizr. Everyone agrees it would be nice, but no one has ever been able to write a test without a significant amount of false positives.

The short version: there are a slew of devices that support touch events *and* mouse events (microsoft surfaces). Users even mix and match these events in single interactions with apps. Detecting support for touch doesn’t in anyways guarantee that touch events are going to occur, and visa versa.

(Also mooost devices lie in interesting ways)

roblarsen:
There's been a lot of discussion here about this very issue. #869 is the most commented issue in the repo. 

If you're unfamiliar with the complexity, this [remains a good place to start](https://patrickhlauke.github.io/getting-touchy-presentation/#Cover) although there are some changes (more complexity!) to the landscape since it was last updated. 

[...]

Issue #2428, opened 2019-03-18
Modernizr touch events on not touch devices


Giovancruz:
Hi, 

I think that the Pull request #2409, to Modernizr is broken my site, because are setting True to Modernizr.touch events on no touch devices as follow:
Old Version:
![image](https://user-images.githubusercontent.com/16118429/54561389-4ead7c00-49a3-11e9-94f1-098a9710a11f.png)

New Version:
![image](https://user-images.githubusercontent.com/16118429/54561422-6422a600-49a3-11e9-8e50-7c4eff6ca4d6.png)

I can't share the site link because is not published yet.

Details:
OS: Ubuntu 16.04 LTS
Browser: Google Chrome v72.0.3626.121 64 bits

Relates to: #2409

rejas:
This is not a bug but a feature @Giovancruz ;-) The test detects if the browser is capable of detecting touches, not if the device is a touch dedevice. See the warning comment in the code:
https://github.com/Modernizr/Modernizr/blob/5afbd1205843ea0c3bb920f073a47efc783533b4/feature-detects/touchevents.js#L12

Issue #2156, opened 2017-02-10
`touchevents` doesn't appear to be working


jchristianhall:
I'm on version 3.3.1 and I recently noticed that every device I've been testing is having `no-touchevents` applied incorrectly. Correct me if I'm wrong, but that test is supposed to detect whether a browser supports the Touch Events spec. However, on the browsers I've tested that should support that spec, they're still showing up as `no-touchevents`. We recently upgraded to the latest Modernizr, so I'm wondering if this introduced a regression.

patrickkettner:
what devices are you testing on?

jchristianhall:
I've tried out iPhone 7 (iOS 10.2), Samsung Galaxy Tab 4 (Android 4.4.2), and a variety of other devices through Browserstack and the iOS simulator.

patrickkettner:
quick [sanity check on codepen](https://codepen.io/anon/pen/dNQmja) is returning touchevents on my iPhone 7 and galaxy tab 5. Could you post a quick dmeo showing the problem?

jchristianhall:
@patrickkettner I'm fairly confident this is a problem with our environment and not Modernizr itself. Thanks for looking into it though!

patrickkettner:
lemme know if youd like any help figuring it out!
cheers

PR #2432, opened 2019-03-22
Proposal: remove touchevents from the project

roblarsen:
This PR removes the `touchevents` detect from the project.

While renaming the test did, in fact, make the test accurately describe what it was testing, it still doesn't do what people think it does. this creates endless confusion (more than we see) and bug reports (the tip of the iceberg of people having problems with this detect.)

We can't create a test that does what people think this test does so we should end the confusion and remove the test. If people want the test in their projects it's an easy one to implement.

Additionally, I would argue that testing for what people _think_ this test does is harmful to users and contributing to that pattern (even via a misunderstanding of what this test does) is bad for the web.

See:

#2431
#2430
#2428
#2134
#548
#1731
#2156
#1628
#1225
#869
#1344
#1179
#884
#1161


rejas:
+1 for this. I'd like to see some form of documentation about this removal in the docs through this PR too.

Additionally: Are there other tests like this that are confusing (because they test the brwoser, not the device) ?

roblarsen:
IIRC all the docs are generated by the feature detects themselves, right? That would be weird here.

rejas:
I thought of the normal README in the github as well as in https://modernizr.com/docs (that page is generated via hbs templates in modernizr-neue too)

roblarsen:
I've got a branch with a docs update for this. I'll open a PR shortly.

scott-thrillist:
https://github.com/Modernizr/Modernizr/commit/ab5febb484f2c8ebfed114459731ba65be20dfa2

Yes, please just remove it at this point. It really stinks that we can't reliably define how the user is interacting with their device so can do things such as preventing clicks from registering as a `hover` event on an iPad, but the more I think about it that's on the browser creators to sort out.

roblarsen:
Sorry, I spaced on the other half of this!

rejas:
NO worries @roblarsen but will you come back to it?

roblarsen:
Sure, yeah. I have a branch with some docs, I don't remember where I was with that, but I'll pull it down and take a look

rejas:
Just a thought: Maybe renaming it to "browsertouch" would be a good middleground (keeping it but indicating that its the brwoser not the device that is tested)?

rejas:
Now officially deprecated with #2472

Issue #2307, opened 2018-04-23
'datachannel'-feature is giving a false-negative on iOS 11.3


nseidel:
We've already tested Web-RTC on iPads with iOS 11.3 and all features were supported, including datachannel. Still, when I check with Modernizr, I get a false result. Any help would be appreciated. See also:
[my question at stack overflow](https://stackoverflow.com/questions/49315459/modernizr-datachannel-feature-with-ios)

DominikTrenz:
We have the same problem. You can fix it by adding a empty string to the domPrefixes so the unprefixed "RTCPeerConnection" is tested for datachannel. Just a quick hack

rejas:
Thx for the info @DominikTrenz and sorry for the late reply @nseidel 
Seems like the datachannel test doesnt check for the unprefixed version:
https://github.com/Modernizr/Modernizr/blob/5afbd1205843ea0c3bb920f073a47efc783533b4/feature-detects/webrtc/datachannel.js#L20

Anyone of you want to take a stab at creating a PR for this?

PR #2437, opened 2019-04-19
fix datachannel #2307

DominikTrenz:
#2307


rejas:
Awesome. Thx for the quick PR

Issue #2433, opened 2019-04-06
Proposal: scrollToOptions detection


asmarcz:
When using `scrollTo` function you can choose between two syntaxes. The scrollToOptions object is not supported in Edge, Safari and IE.

`scrollTo` scrolls to 0, 0 when it is supplied "invalid arguments" so the support can be detected by this.

I don't know whether this is something worth being in Modernizr. I would appreciate any help with my first contribution.

rejas:
hi @asmarcz this could be indeed something for modernizr. what you need help with foremost? on how to write a PR in general or on how to write a test for modernizr?

asmarcz:
@rejas Thank you for response. I will look into it in two days. I will ask here if I get stuck.

asmarcz:
I have created test file in feature-detects folder and registered it in config-all.json. Then I generated bundle using `./bin/modernizr -c lib/config-all.json` however when I wanted to try out whether it works, I found out that there is an error in the bundle itself.

`modernizr.js:6866 Uncaught TypeError: Cannot read property 'removeChild' of null
    at Object.fn (modernizr.js:6866)
    at testRunner (modernizr.js:144)
    at modernizr.js:10169
    at modernizr.js:10188`

What should I do? Should I clone a different branch or something?

rejas:
Without looking at the cdoe I cant say anything for sure. Could you create a PR from your branch so I can have a look?

asmarcz:
@rejas Turned out I misused Modernizr, my bad. I created a pull request. I tried to understand the Travis log but could not wrap my head around it. Everything seems to work OK when I build and test locally. #2438

PR #2438, opened 2019-04-22
scrollToOptions dictionary feature detection

asmarcz:
Fixes https://github.com/Modernizr/Modernizr/issues/2433


rejas:
hi @asmarcz sorry for the long pause. could you rebase your PR against the latest master so that we can see if the tests dont fail anymore?

feature-detects/scroll-to-options.js
```diff
@@ -0,0 +1,27 @@
+/*!
+{
+  "name": "scrollToOptions dictionary",
+  "property": "scrolltooptions"
+}
```
rejas: coudl you add at least the tags for author (yourself :-) to this? also if there is a caniuse property check for this add this too as well as other links that might be useful.

lib/config-all.json
```diff
@@ -253,6 +253,7 @@
     "requestanimationframe",
     "script/async",
     "script/defer",
+    "scroll-to-options",
```
rejas: could you move the test into the css directory? I think its more fitting there

asmarcz: @rejas  Why would you move it into the css directory? It is a test of JS function and its behaviour. Sorry for the pause, didn't read GitHub notifications.

feature-detects/scroll-to-options.js
```diff
@@ -0,0 +1,27 @@
...
+!*/
+define(['Modernizr'], function (Modernizr) {
```
rejas: Please use helper classes like this:

define(['Modernizr', 'createElement', 'getBody'], function (Modernizr, createElement, getBody) {

feature-detects/scroll-to-options.js
```diff
@@ -0,0 +1,27 @@
...
+  Modernizr.addTest('scrolltooptions', function () {
+    var returnTo = window.pageYOffset;
+    var needsFill = document.body.clientHeight <= window.innerHeight;
```
rejas: so now use
    var body = getBody();
    var needsFill = body.clientHeight <= window.innerHeight;

(and for the other occurences of the body element too

feature-detects/scroll-to-options.js
```diff
@@ -0,0 +1,27 @@
...
+    if (needsFill) {
+      var div = document.createElement('div');
```
rejas: same here, use helper class 
      var div = createElement('div');

lib/config-all.json
rejas: you're right, forget that :-)

feature-detects/scroll-to-options.js
```diff
@@ -0,0 +1,27 @@
+/*!
+{
+  "name": "scrollToOptions dictionary",
+  "property": "scrolltooptions"
+}
+!*/
```
rejas: some docs on what the test does and expects would be very helpful

rejas:
@asmarcz still working on the requested changes? please also make sure that the test dont fail

asmarcz:
No, sorry. I don't plan to waste another time with this time scattered issue.

Issue #1805, opened 2015-12-10
vh unit issue on ios safari


jimbatamang:
Even though the latest ios safari supports vh units, modernizr is passing negative class "no-cssvhunit".

patrickkettner:
just for historical sake, latest is 9.0.2, right?

alrra:
> latest is 9.0.2, right?

@patrickkettner [`9.2.0`](https://support.apple.com/kb/DL1842?locale=en_US).

jimbatamang:
@patrickkettner, @alrra : I am looking at iPhone 6, version 9.1

hexalys:
The test will fail in iOS 8+ because 100vh is the innerHeight + the bottom interactive bar. What happens is that when the bar disappear, you get a true 100vh. But with the initial bar, the test is justifiably negative.

I managed to detect the bar condition (and height to some extend) in my own custom script, in order to assess when to correct vh units subtracting the bar height. But technically, the current test works as intended by detecting a bug. It's a tough call because it's a design decision on Apple's part. That issue occurs on the Firefox OS Phone as well. See [this article](http://nicolas-hoizey.com/2015/02/viewport-height-is-taller-than-the-visible-part-of-the-document-in-some-mobile-browsers.html) by Nicolas Hoizey for details on the issue.

Sadly the only way to use vh correctly accross an iOS view is to rely detection heuristics of that bar.

nhoizey:
Thanks @hexalys for mentioning my post about this `100vh` not being the real full height on iOS and Firefox OS (at least).

This is really annoying for the development of a full screen app, for example a game like my own https://github.com/nhoizey/esviji

Here is the welcome screen iOS users currently get… :-/
![img_6084](https://cloud.githubusercontent.com/assets/78213/12019230/ce84c10c-ad6d-11e5-81ea-9c8cc919584d.PNG)

nhoizey:
It seems something add been done for iOS8 in #1430, maybe it can be extended for iOS9…

hexalys:
@nhoizey It's quite frankly impossible to adjust the test with a fixed set of code to reflect iOS support. The support varies according to window conditions, and would require to update 'cssvhunit' or 'no-cssvhunit' dynamically. e.g. In portrait, the bar at the bottom is 49px, while it's 44 at the top in landscape.

[...]

PR #2452, opened 2019-05-24
Fix issue #1805

Minilfat:
That issue is actual not only for iOS devices but Android browser as well. I wrote more detailed [here](https://github.com/Modernizr/Modernizr/issues/1805#issuecomment-495656198). 
Idea how to fix this issue was suggested [here](https://github.com/Modernizr/Modernizr/issues/1805#issuecomment-284103672).

### DIFF
--- a/feature-detects/css/vhunit.js
+++ b/feature-detects/css/vhunit.js
@@ -14,9 +14,9 @@
   }]
 }
 !*/
-define(["Modernizr", "testStyles", "computedStyle"], function(Modernizr, testStyles, computedStyle) {
-  testStyles("#modernizr { height: 50vh; max-height: 10px; }", function(elem) {
-    var compStyle = parseInt(computedStyle(elem, null, "height"), 10);
-    Modernizr.addTest("cssvhunit", compStyle === 10);
+define(['Modernizr', 'testStyles', 'computedStyle'], function(Modernizr, testStyles, computedStyle) {
+  testStyles('#modernizr { height: 50vh; max-height: 10px; }', function(elem) {
+    var compStyle = parseInt(computedStyle(elem, null, 'height'), 10);
+    Modernizr.addTest('cssvhunit', compStyle == 10);
   });
 });
--- a/feature-detects/css/vhunit.js
+++ b/feature-detects/css/vhunit.js
@@ -14,7 +14,7 @@
   }]
 }
 !*/
-define(['Modernizr', 'testStyles', 'computedStyle', 'roundedEquals'], function(Modernizr, testStyles, computedStyle, roundedEquals) {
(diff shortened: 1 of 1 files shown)


rejas:
Thx for the PR but I got a quetion: This basically reverts https://github.com/Modernizr/Modernizr/pull/2264 which tried to fix https://github.com/Modernizr/Modernizr/issues/1569 
SO my question is: how can we get a solution that dixes both problems?

Minilfat:
> Thx for the PR but I got a quetion: This basically reverts #2264 which tried to fix #1569
> SO my question is: how can we get a solution that dixes both problems?

Ohh, I was not aware about those issues. I will let you know if we come up with some solution. Thank you

Minilfat:
> Thx for the PR but I got a quetion: This basically reverts #2264 which tried to fix #1569
> SO my question is: how can we get a solution that dixes both problems?

@rejas  hello. What do you think about this solution: we set the height of the element to 50vh, but also set "max-height" to be 10px. In case the "vh" is supported by browser then the height of the element is 10px.

The code is following: 
```javascript
define(["Modernizr", "testStyles", "computedStyle"], function(Modernizr, testStyles, computedStyle) {
  testStyles("#modernizr { height: 50vh; max-height: 10px; }", function(elem) {
    var compStyle = parseInt(computedStyle(elem, null, "height"), 10);
    Modernizr.addTest("cssvhunit", compStyle === 10);
  });
});
```

feature-detects/css/vhunit.js
```diff
@@ -14,11 +14,9 @@
...
 !*/
-define(['Modernizr', 'testStyles', 'computedStyle', 'roundedEquals'], function(Modernizr, testStyles, computedStyle, roundedEquals) {
-  testStyles('#modernizr { height: 50vh; }', function(elem) {
-    var height = parseInt(window.innerHeight / 2, 10);
+define(['Modernizr', 'testStyles', 'computedStyle'], function(Modernizr, testStyles, computedStyle) {
+  testStyles('#modernizr { height: 50vh; max-height: 10px; }', function(elem) {
     var compStyle = parseInt(computedStyle(elem, null, 'height'), 10);
-
-    Modernizr.addTest('cssvhunit', roundedEquals(compStyle, height));
+    Modernizr.addTest('cssvhunit', compStyle == 10);
```
rejas: why == instead of === ?

Minilfat: Both of them have type "Number", don't they? I can change to '===' not a problem at all. Should I do it?

rejas: === is in 99 % of the time preferable even if you think that the types are always the same...

Minilfat: Fixed! :)

PR #2453, opened 2019-05-27
Added feature detect native lazy loading img

horprogs:
Detects, that image supports tag `loading` (native lazy loading). Returns true if exists property `loading` in `HTMLImageElement.prototype`. Otherwise, returns false.

- Chrome Version 76.0.3805.0 (Official Build) canary (64-bit) with flag #enable-lazy-image-loading (result: true)
- Chrome Version 74.0.3729.169 (Official Build) (64-bit) (result: false)
- Safari 12.1.1 (13607.2.6.1.2) (result: false)


feature-detects/img/loading.js
```diff
@@ -0,0 +1,19 @@
+/*!
+{
+  "name": "image native lazy loading",
+  "property": "loading",
+  "caniuse": "loading",
```
rejas: the caniuse property "loading-lazy-attr" (https://caniuse.com/#feat=loading-lazy-attr)

feature-detects/img/loading.js
```diff
@@ -0,0 +1,19 @@
+/*!
+{
+  "name": "image native lazy loading",
```
rejas: better: image and iframe native lazy loading

feature-detects/img/loading.js
```diff
@@ -0,0 +1,19 @@
...
+  "property": "loading",
```
rejas: I'd like to use "lazyloading" for this (which you also use in the actual test down below)

feature-detects/img/loading.js
```diff
@@ -0,0 +1,19 @@
...
+  "caniuse": "loading",
+  "tags": ["image", "lazy", "loading"],
+  "notes": [{
+    "name": "Native image lazy-loading for the web",
+    "href": "https://addyosmani.com/blog/lazy-loading/"
+  }]
+}
+!*/
+/* DOC
+Test for the loading attribute of images
```
rejas: ... and iframes

horprogs:
Thx you, I fixed it

lib/config-all.json
```diff
@@ -204,6 +204,7 @@
     "img/crossorigin",
     "img/jpeg2000",
     "img/jpegxr",
+    "img/loading",
```
rejas: just one more thing: rename the file from just "loading" to "lazyloading" Makes it clearer what the feature does and correpsonds to the property too.

horprogs: No problem. Renamed.

rejas:
Thx a lot!

PR #2459, opened 2019-06-24
Add proxy detection

BrockBeaudry:
* Adds detection for the Proxy object

https://caniuse.com/#feat=proxy

### DIFF
--- a/feature-detects/proxy.js
+++ b/feature-detects/proxy.js
@@ -3,7 +3,7 @@
   "name": "Proxy Object",
   "property": "Proxy",
   "caniuse": "proxy",
-  "tags": ["dom"],
+  "authors": ["Brock Beaudry"],
   "notes": [{
     "name": "MDN Docs",
     "href": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy"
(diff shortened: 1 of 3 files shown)


feature-detects/proxy.js
```diff
@@ -0,0 +1,21 @@
+/*!
+{
+  "name": "Proxy Object",
+  "property": "Proxy",
+  "caniuse": "proxy",
```
rejas: you can add a authors tag here with you name/userhandle if you want :-)

feature-detects/proxy.js
```diff
@@ -0,0 +1,21 @@
...
+  "tags": ["dom"],
```
rejas: might be an idea to put it in the dom directory if it has that tag?

feature-detects/proxy.js
BrockBeaudry: Added, thanks for the suggestion

feature-detects/proxy.js
BrockBeaudry: I don't think it actually has this tag so I went ahead and removed it

rejas:
Thx for the PR.

PR #2466, opened 2019-08-26
Set classes for inputtypes (#2130)

### LINKED ISSUE #2130: Modernizr.inputtypes no class set.
Tested this in latest Chrome, Safari, and Firefox. No class set for any input types.

backflip:
This change will make sure that `setClasses` works for every `inputtypes` test.
Fixes #2130

### DIFF
--- a/test/browser/integration/classes.js
+++ b/test/browser/integration/classes.js
@@ -27,7 +27,7 @@ describe('classes', function() {
               name = name.replace('-', '');
               expect(Modernizr[name]).to.be(result);
             } else {
-              name = name.split('-');
+              name = name.split(/-(.+)/); // split at first occurence of '-'
               expect(Modernizr[name[0]]).to.not.be(undefined);
               expect(!!Modernizr[name[0]][name[1]]).to.equal(result);
             }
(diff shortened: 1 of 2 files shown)


test/browser/integration/classes.js
```diff
@@ -27,7 +27,7 @@ describe('classes', function() {
               name = name.replace('-', '');
               expect(Modernizr[name]).to.be(result);
             } else {
-              name = name.split('-');
+              name = name.split(/-(.+)/); // split at first occurence of '-'
```
backflip: `inputtypes` has a `datettime-local` property. Without this change, we would split the name `inputtypes-datettime-local` into three instead of two parts.

rejas:
Thx for the PR but this would add new js tests too (so also inputtypes-datetime while currently we have the object inputtypes.datetime.
Maybe just adding the code to add classes would suffice? Something like this:
`setClasses([(!!bool && bool !== false ? '' : 'no-') + 'inputtypes-' + props[i]]);`

Besides, I always wondered why some tests had been put into objects, I dont see the reason why except having them in one place but maybe @patrickkettner can shed some light on this?

backflip:
> Maybe just adding the code to add classes would suffice?

Wouldn't this set the classes even if the `setClasses` option was `false`?

rejas:
Thx for your initial work. I took it and my comments to provide a new PR that also fixes some other tests: #2469

backflip:
Awesome, looking forward to this getting merged.

PR #2470, opened 2019-09-08
Fix unicode detection in latest chrome

rejas:
Fixes #2470 by using a wider unicode glyph for detection. 
The current star seems to have the same width in chrome (or isnt supported anymore)


rejas:
Closing since it seems to be a chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1002338

PR #2471, opened 2019-09-10
Check if userAgent is empty

listton:
Some browsers allow to have empty userAgent. Therefore, we need to check ua before using "indexOf" on it.


rejas:
Thx for the PR. Could you specify in it too which browsers for example are doing this?

listton:
> Thx for the PR. Could you specify in it too which browsers for example are doing this?

It is certainly possible in different versions of Chrome, Firefox, Opera and Edge. There are many articles "How to change user agent" and "How to remove user agent", for example: https://www.maketecheasier.com/change-user-agents-chrome-firefox-edge/

Issue #2468, opened 2019-09-05
Unicode detection in Chrome 76


liviascapin:
I'm using Google Chrome version 76.0.3809.132 (64-Bit) and modernizr 3.6.0 with Unicode-Character detection (https://modernizr.com/download/?-unicode-setclasses)

And modernizr is telling me that unicode is not 'supported'.
I tried debugging a bit and checked `missingGlyph.offsetWidth` and `star.offsetWidth` and they really seem to be the same now... I tried another unicode symbol (I randomly picked `&#2342;`) and this one seems to have a different offsetWidth.

Funnily enough I use the same chrome version on my Android and it does not seem to have the same problem, but I reproduced it on multiple (windows) computers.

I think for now I will just remove this check, but I just wanted to let you know that there might be a problem. Sadly I don't know how to fix it.

shockthetoast:
We have run into the exact same issue. Our current solution is also to remove the unicode checks - hopefully a browser that passes any of the others has unicode support.

It worked fine in 76.0.3809.100, and failed for the first time (on release) in 76.0.3809.132.

I actually ran into this for the first time when testing the developer preview of Microsoft Edge, which is switching to use Chromium for their rendering engine. When it failed there I did a little digging and ran into the same offsetWidth issues as @liviascapin did. However, I naively thought it was just an issue Microsoft had introduced. It turns out that Edge must have been using a pre-release version of Chromium with this change. I submitted an issue in Edge, but I did so using its feedback reporting tools that basically send info into the aether, so of course I never received a response.

It's interesting to note that this release was supposed to only be to fix some security bugs. So either something else slipped in, or one of the security fixes affected rendering.

rejas:
Thx for your inout you two. I tried to fix this in #2470 could you try it out and report back if that fixes it for you too?

shockthetoast:
@rejas It's still failing for us. Interestingly, if I add some console output, Chrome shows the offsetWidth for both characters as 2250. If I run it in Firefox, I get the offset widths as 883 for missingGlyph and 1820 for the unicode comet character.

My first thought was that Chrome is setting some maximum offsetWidth, but when comparing with Firefox it's reporting an offsetWidth many times larger - more than seems to be reasonably explained by quirks in the browsers' font handling differences.

[...]

PR #2472, opened 2019-09-21
Proposal: Deprecate some tests

rejas:
This PR marks these tests as deprecated and to be removed with v3.9;

Unicode (https://github.com/Modernizr/Modernizr/issues/2468) has dubious value 
Touchevents (https://github.com/Modernizr/Modernizr/pull/2432) is often misinterpreted


roblarsen:
This might be a better solution that outright removing the tests. `console.warn`?

rejas:
Doy ou mean a warning when building or running the tests?
As for removing: Mabye a major release (4.0) would be better, since those can mean breaking changes per semver definition?

roblarsen:
Building would be less intrusive and would target the developer using the tests. 

And yeah, if they're removed a major release would be the way to go.

README.md
```diff
@@ -104,6 +104,14 @@ http://localhost:8080/test/unit.html
 http://localhost:8080/test/integration.html
 ```
  
+## Deprecation
+
+These tests are considered deprecated. They are not included anymore in the default build
+and will re removed in the next major version:
```
paulirish:
```suggestion
and will be removed in the next major version:
```

rejas:
After looking at the code, I am not the wiser where to add a warning when building modernizr...

Anyway, merging this for 3.8 would be a thing and adding a better warning one for 3.9 :-) Any objections?

PR #2475, opened 2019-10-03
Add support for ShadowRoot DOM api feature detect

pascalim:
Add support for ShadowRoot DOM api feature detect through `shadowroot` and `shadowroot-legacy` tests

### DIFF
--- a/feature-detects/dom/shadowroot.js
+++ b/feature-detects/dom/shadowroot.js
@@ -1,0 +1,20 @@
+/*!
+{
+  "name": "Shadow DOM API",
+  "property": "shadowroot",
+  "caniuse": "shadowdomv1",
+  "notes": [{
+    "name": "MDN Docs",
+    "href": "https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot"
+  }],
+  "authors": ["Kevin Coyle (@kevin-coyle-unipro)", "Pascal Lim (@pascalim)"],
+  "tags": ["dom"]
+}
+!*/
+
+/* DOC
+Detects support for the Shadow DOM API.
+*/
+define(['Modernizr', 'createElement'], function (Modernizr, createElement) {
+  Modernizr.addTest('shadowroot', 'attachShadow' in createElement('div'));
+});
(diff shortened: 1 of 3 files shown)


rejas:
Thx for the PR. But it test for both the old and the newer ShadowDom Api. The first check is against the deprecated call: https://caniuse.com/#feat=mdn-api_element_createshadowroot
while the later is the favored modern one.
Therefor I guess the test should be split into two, what dou you think?

pascalim:
Good eye, yep sounds good, updated PR with `shadowroot` and `shadowroot-legacy` tests

rejas:
Tests are failing, did you notice? Please fix that first  :-)

pascalim:
@rejas it's a case of me not reading the documentation first :p Ok refactored the tests and it seems to appease the test runners

feature-detects/dom/shadowroot.js
```diff
@@ -0,0 +1,22 @@
+/*!
+{
+  "name": "Shadow DOM API",
+  "property": ["shadowroot", "shadowrootlegacy"],
```
rejas: please split up again in two files, that way we can get rid of one if we want to deprecate the old one

feature-detects/dom/shadowroot.js
```diff
@@ -0,0 +1,22 @@
...
+  "notes": [
+    {
+      "name": "MDN Docs",
+      "href": "https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot"
+    }
+  ],
+  "authors": ["Kevin Coyle", "Pascal Lim"],
```
rejas: maybe add your github usernames too: "Pascal Lim (@pascalim)"

feature-detects/dom/shadowroot.js
```diff
@@ -0,0 +1,22 @@
...
+  "tags": ["dom"]
+  }
```
rejas: please add the caniuse references too https://caniuse.com/#feat=shadowdom https://caniuse.com/#feat=shadowdomv1

pascalim:
PR updated with requested changes

rejas:
thx a lot! good job!

PR #2478, opened 2019-10-30
[Modernizr]: Support to disable openDatabase driver for testing

ossdev07:
Providing support to test if openDatabase driver is virtually removed.
In some cases users would like to test what if openDatabase driver doesn't support by setting openDatabase to null. In this case the test application/method will fail to use openDatabase driver and Modernizr must behave same.

Signed-off-by: ossdev07 <ossdev@puresoftware.com>

### DIFF
--- a/feature-detects/storage/websqldatabase.js
+++ b/feature-detects/storage/websqldatabase.js
@@ -9,5 +9,12 @@
 define(['Modernizr'], function(Modernizr) {
   // Chrome incognito mode used to throw an exception when using openDatabase
   // It doesn't anymore.
-  Modernizr.addTest('websqldatabase', 'openDatabase' in window);
+  Modernizr.addTest('websqldatabase', function() {
+    if (window.openDatabase === null) {
+      return false;
+    }
+    else {
+      return 'openDatabase' in window;
+    }
+  });
 });


ossdev07:
Ping @rejas, could you please have a look at this PR?

rejas:
Hi @ossdev07 
I dont see a good reason to add this to the library yet. Why should a user set the openDatabase to null and then test again with Modernizr? He already seems to know what happens so no need to teach Modernizr any new edge cases...

ossdev07:
Hi Rejas,
The application developer would like to deploy their application for multiple browsers out of which all browsers may not support openDatabase driver and this must have been handled accordingly. let’s say a developer has automated his test cases using a browser that supports openDatabase and he wants to write a test case to see the behaviour of application for those browsers which doesn't support openDatabase. This PR will help the application developer to deal with this scenario.

rejas:
Your scenario is not logical in my opinion. If you test Browser A by setting this feature undefined in Browser B why not test Browser A directly? Otherwise you cannot be sure if the tests are correct since Browser B might have other differences to Browser A that your tests dont take into account. 
Otherwise going by your logic we would have to implement fallbacks for these type of scenarios into all of our tests.

ossdev07:
@rejas, Here the only point is, developing automation tests for multiple browser will result in larger test execution time with less benefit and all the feature supported by one browser is not necessary to support in another.

rejas:
Modernizr is about testing the browser. If you want to fake testing another browser, why test at all?
Sorry, wont merge this unless you can show me if there is already something like this in the code.

PR #2487, opened 2019-12-16
feature detects for ES6 Symbol;

buhichan:
Adds feature detects for ES6 Symbol. well-known symbols which are not part of es6 such as asyncIterator are not included.

### DIFF
--- a/feature-detects/es6/symbol.js
+++ b/feature-detects/es6/symbol.js
@@ -2,6 +2,7 @@
 {
   "name": "ES6 Symbol",
   "property": "es6symbol",
+  "caniuse": "mdn-javascript_builtins_symbol",
   "notes": [{
     "name": "Official ECMAScript 6 specification",
     "href": "https://www.ecma-international.org/ecma-262/6.0/#sec-symbol-constructor"
@@ -27,11 +28,10 @@ define(['Modernizr'], function(Modernizr) {
     Symbol.match &&
     Symbol.prototype &&
     Symbol.replace &&
-    Symbol.search && 
+    Symbol.search &&
     Symbol.species &&
     Symbol.split &&
-    Symbol.toPrimitive && 
-    Symbol.toStringTag && 
+    Symbol.toPrimitive &&
+    Symbol.toStringTag &&
     Symbol.unscopables));
 });
-  
--- a/feature-detects/es6/symbol.js
+++ b/feature-detects/es6/symbol.js
@@ -1,0 +1,37 @@
+/*!
+{
+  "name": "ES6 Symbol",
+  "property": "es6symbol",
+  "notes": [{
+    "name": "Official ECMAScript 6 specification",
+    "href": "https://www.ecma-international.org/ecma-262/6.0/#sec-symbol-constructor"
+  },{
+    "name": "MDN web docs",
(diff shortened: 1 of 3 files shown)


feature-detects/es6/symbol.js
```diff
@@ -0,0 +1,34 @@
+/*!
+{
+  "name": "ES6 Symbol",
+  "property": "es6symbol",
+  "notes": [{
```
rejas: maybe add a link to the mdn too 
 https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Symbol

feature-detects/es6/symbol.js
```diff
@@ -0,0 +1,34 @@
...
+    "name": "Official ECMAScript 6 specification",
+    "href": "https://www.ecma-international.org/ecma-262/6.0/#sec-symbol-constructor"
+  }],
+  "polyfills": ["es6shim"],
+  "authors": ["Ron Waldon (@jokeyrhyme)"],
```
rejas: you can put your name here :-)

feature-detects/es6/symbol.js
```diff
@@ -0,0 +1,34 @@
+/*!
+{
+  "name": "ES6 Symbol",
+  "property": "es6symbol",
+  "notes": [{
+    "name": "Official ECMAScript 6 specification",
+    "href": "https://www.ecma-international.org/ecma-262/6.0/#sec-symbol-constructor"
+  }],
+  "polyfills": ["es6shim"],
```
rejas: are you sure that polyfill is covering all of symbol too?

feature-detects/es6/symbol.js
```diff
@@ -0,0 +1,34 @@
...
+  "authors": ["Ron Waldon (@jokeyrhyme)"],
+  "tags": ["es6"]
+}
```
rejas: you can also add a caniuse tag here : https://caniuse.com/#feat=mdn-javascript_builtins_symbol_for

feature-detects/es6/symbol.js
buhichan: I thought here are names of the authors of polyfills, and I'm not one.

feature-detects/es6/symbol.js
buhichan: I'm not sure where to look for these tags, all other es6 features' tags are just "es6". Maybe ["es6","symbol"] is ok ?

feature-detects/es6/symbol.js
buhichan: My mistake, es6shim does not cover Symbol.   
Will add this polyfill: https://github.com/medikoo/es6-symbol.
How do you decide which polyfill implementation is the best for this feature?

feature-detects/es6/symbol.js
buhichan: Added.

feature-detects/es6/symbol.js
buhichan: Changed.

feature-detects/es6/symbol.js
buhichan: changed to ["es6","symbol"]

buhichan:
thanks for replying. updated.

rejas:
Thx for the PR and soon a happy new year :-)

PR #2513, opened 2020-03-12
ES6/Rest parameters

dabretin:
I didn't find any detection of rest parameters and I think it's very important because it could break script load if not supported...
So another try of PR :)

PS: widely inspired of `es6/arrow`


dabretin:
Ouppsssss...
Sorry, the test in this PR is not for ES6 rest but for ES6 deconstruct rest...

PR #2515, opened 2020-03-13
ES7 array test (includes method)

### LINKED ISSUE #2376: FR: Detection for Array.prototype.includes()
I don't see any feature detection in Modernizr for [`Array.prototype.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes). I would find this useful, would others?

Perhaps we could add this to the [series of checks in es6/array.js](https://github.com/Modernizr/Modernizr/blob/v3.6.0/feature-detects/es6/array.js#L20-L28), or is `Array.prototype.includes()` considered post-es6?

Modernizr is an excellent library - a staple of modern web development. Thanks for everyone who has worked to make it so great!

dabretin:
Fixes #2376 

I hope this one will be ok :)
ES7 introduce a new method (includes) for array. This PR check the presence of this method on the array prototype.

PS: widely inspired of es6/array

### DIFF
--- a/feature-detects/es7/array.js
+++ b/feature-detects/es7/array.js
@@ -14,7 +14,7 @@
 /* DOC
 Check if browser implements ECMAScript 7 Array per specification.
 */
-define(["Modernizr"], function(Modernizr) {
-	Modernizr.addTest("es7array", !!(Array.prototype &&
+define(['Modernizr'], function(Modernizr) {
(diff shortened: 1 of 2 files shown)


rejas:
Looks good, will merge now :-)

dabretin:
Thx! :+1:

PR #2518, opened 2020-03-13
ES6 - String template

dabretin:
Check ES6 string template support (``const a = `Hello ${name}`;``).
I didn't modify `es6/string` because there is no shim for that...


feature-detects/es6/string-template.js
```diff
@@ -0,0 +1,25 @@
+/*!
+{
+  "name": "ES6 Template Strings",
+  "property": "stringtemplate",
+  "notes": [{
+    "name": "ECMAScript 6 draft specification",
+    "href": "https://tc39wiki.calculist.org/es6/template-strings/"
+  }],
+  "authors": ["dabretin"],
+  "tags": ["es6"]
```
rejas: please add a caniuse tag https://caniuse.com/#feat=template-literals

feature-detects/es6/string-template.js
```diff
@@ -0,0 +1,26 @@
...
+}
+!*/
+/* DOC
+Check if browser implements ECMAScript 6 String template.
+*/
+define(['Modernizr'], function(Modernizr) {
+  Modernizr.addTest('stringtemplate', function() {
+    try {
+      // eslint-disable-next-line
+      return eval('(function(){var a=1; return `-${a}-`;})()') == '-1-';
```
rejas: shouldnt this be a strict comparison `===` ?

dabretin: I don't think it's necessary ('-1-' can only be a string) but your're right: never too many precautions !

PR #2457, opened 2019-06-11
Add Av1 codec detection

MarkelFe:
Completely not tested AV1 detection code.

Possible problems:
- AV1 can be contained by a lot of containers .mkv `video/x-matroska`, webm `video/webm`, .mp4 `video/mp4` and not sure if any other, research is required. Right now is only being tested for .mp4.
- AV1.1 could be a problem, not completely sure.

It simply needs some research and testing 🙃

PR #2534, opened 2020-04-10
Add av1 and h265 codec detection

### LINKED ISSUE #2231: H265 check
It would be nice to have H265 support check in Modernizr. Edge and Safari already support HEVC.

MarkelFe:
**Similar pull requests**: #2457 (mine)
Closes #2231 

**Changes**:
- Add av1 codec detection
- Add x265 codec detection
- Added some missing subproperties to the description on the web

**Comments**:
No new tests needed to be created as the test scripted already check for all existing codecs. I've tested with Firefox 75 and Edge 18 (with and without [hevc extensions](https://www.microsoft.com/en-us/p/hevc-video-extensions-from-device-manufacturer/9n4wgh0z6vhq?activetab=pivot:overviewtab) installed) in gh-pages and they match [caniuse](https://caniuse.com/#feat=av1). Furthermore `npm test` shows 2 new tests passed.

### DIFF
--- a/feature-detects/video.js
+++ b/feature-detects/video.js
@@ -18,7 +18,7 @@
 /* DOC
 Detects support for the video element, as well as testing what types of content it supports.
 
-Subproperties are provided to describe support for `ogg`, `h264` and `webm` formats, e.g.:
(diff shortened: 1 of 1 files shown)


MarkelFe:
Just found out that this would solve #2231

PR #2538, opened 2020-04-21
Fixes an issue with AVG Secure browser throwing exceptions

Sembiance:
When using either the AVG or AVAST Secure Browser, if you turn "Anti-Fingerprinting" on, it will throw exceptions when attempting to use the result of .toDataURL() from a canvas. This causes 'test/canvas/todataurl' (feature-detects/canvas/todataurl.js) to fail which breaks Modernizr.

The solution is to wrap the test in a try/catch, just like you were already doing for todataurlwebp for firefox anti-fingerprinting but also for todataurlpng and todataurljpeg.


rejas:
Thx for the PR.

PR #2539, opened 2020-04-22
Add avif detection

MarkelFe:
Based on the jpeg2000 test I've added avif detection. 

However, it is a draft pull request for a simple reason: [there is no browser](https://caniuse.com/#feat=avif) to test if it works correctly 😅. I will mark it ready for review whenever there is a browser that supports avif, developments are already being made in [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1625363) and [Chrome](https://chromium-review.googlesource.com/c/chromium/src/+/2101317). At least the false works correctly in Firefox 75 and Brave (Chrome 81). 

Also, I finally learnt how to add tests :)

### DIFF
--- a/feature-detects/img/avif.js
+++ b/feature-detects/img/avif.js
@@ -6,6 +6,7 @@
   "caniuse": "avif",
   "tags": ["image"],
   "authors": ["Markel Ferro (@MarkelFe)"],
+  "polyfills": ["avifjs"],
   "notes": [{
     "name": "Avif Spec",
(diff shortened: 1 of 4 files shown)


MarkelFe:
[Firefox 77 introduces avif support if a flag is activated.](https://caniuse.com/#feat=avif) I obviously downloaded the latest version to check if the feature detect works and...
It works

So with the feature detect tested I'm marking this as ready for review. The [raw output is also available](https://pastebin.com/6MQT84S0) if you desire :)

PR #2545, opened 2020-04-27
Upgrade mkdirp dependency, remove node8 support

### LINKED ISSUE #2544: Upgrade mkdirp to latest version, drop node8 support.
Node8 is not supported anymore, this is now the time to drop it here too by upgrading to the v1 branch of mkdirp (which requires node10)

rejas:
Fixes #2544

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,9 +95,6 @@
     "update-changelog": "auto-changelog --commit-limit false --package",
     "update-license": "node scripts/generate-license.js && git add LICENSE"
   },
-  "directories": {
-    "test": "test"
-  },
   "engines": {
     "node": ">=10"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -95,6 +95,12 @@
     "update-changelog": "auto-changelog --commit-limit false --package",
     "update-license": "node scripts/generate-license.js && git add LICENSE"
   },
+  "directories": {
+    "test": "test"
```
patrickkettner, line 99: any reason?

rejas: just alphabetical order. for what tool is that useful bytheway?

patrickkettner: oh, I didn't realize this was here in the first place.

looks like it was added for grunt back in the day. it did what [you did here](https://github.com/Modernizr/Modernizr/blob/master/gulpfile.babel.js#L16-L30)

id say chuck it, while we are here

rejas: chucked it!

patrickkettner: image

patrickkettner:
:shipit:

PR #2547, opened 2020-04-27
Add metadata documentation

MarkelFe:
I found out that there was no documentation for the metadata of the files, so I created a little Markdown file explaining what every piece of metadata represents + putting all of them together, as all the detects don't have the same amount of metadata.

I didn't know where to put this information, so I created a new file, feel free to suggest any other location.

There is also some of them that I don't know their functionality, I'll comment them on line comments.

### DIFF
--- a/metadata.md
+++ b/metadata.md
@@ -38,7 +38,7 @@ Here it would go a description of the feature test. You can use **markdown** her
 | `property`       | required | The property name established in `Modernizr.addTest` |                   It must be lowercase, without any punctuation                  |
 | `tags`           | optional |    A group that encapsulates many feature detects    |                                                                                  |
 | `caniuse`        | optional |      A conversion table of caniuse and Modernizr     |      Consider adding it to [caniuse.js](test/browser/integration/caniuse.js)     |
-| `authors`        | optional |          List of contributors of the script          |                     There are a couple of ways to express it                     |
+| `authors`        | optional |          List of contributors of the script          |                                                                                  |
 | `builderAliases` | optional |     Used by CI and the web when tests are renamed    |                         Should not be needed in new tests                        |
 | `polyfills`      | optional |     Available polyfills for not working versions     | Any polyfill listed needs to be included in [polyfills.json](lib/polyfills.json) |
 | `aliases`        | optional |   Used if a feature has more than a canonical name   |                             Legacy only - do not use                             |
--- a/metadata.md
+++ b/metadata.md
@@ -39,7 +39,7 @@ Here it would go a description of the feature test. You can use **markdown** her
 | `tags`           | optional |    A group that encapsulates many feature detects    |                                                           |
 | `caniuse`        | optional |      A conversion table of caniuse and Modernizr     | Consider adding it to test/browser/integration/caniuse.js |
 | `authors`        | optional |          List of contributors of the script          |          There are a couple of ways to express it         |
-| `builderAliases` | optional |                                                      |                                                           |
+| `builderAliases` | optional |     Used by CI and the web when tests are renamed    |                                                           |
 | `polyfills`      | optional |     Available polyfills for not working versions     |               Add them in lib/polyfills.json              |
 | `aliases`        | optional |         Other names that the feature goes by         |                                                           |
 | `async`          | optional |       If the test supports async functionality       |                     Defaults to false                     |
--- a/metadata.md
+++ b/metadata.md
@@ -32,17 +32,17 @@ Here it would go a description of the feature test. You can use **markdown** her
 > There is no set order in which they must appear, but it is common to see `name` and `property` at the top while `notes` at the bottom.
 
 ## Item description
-|                  | Necesity |                      Description                     |                        Notes                       |
-|------------------|:--------:|:----------------------------------------------------:|:--------------------------------------------------:|
-| `name`           | required |             Name of the feature detection            |                                                    |
-| `property`       | required | The property name established in `Modernizr.addTest` | Preferably just lowercase, without any punctuation |
-| `tags`           | optional |                                                      |                                                    |
-| `caniuse`        | optional |                                                      |                                                    |
-| `authors`        | optional |          List of contributors of the script          |      There are a couple of ways to express it      |
-| `builderAliases` | optional |                                                      |                                                    |
-| `polyfills`      | optional |     Available polyfills for not working versions     |           Add them in lib/polyfills.json           |
-| `aliases`        | optional |         Other names that the feature goes by         |                   Used for search                  |
-| `async`          | optional |       If the test supports async functionality       |                  Defaults to false                 |
-| `warnings`       | optional |        Notes to the developer using the script       |           Don't mistake it for knownBugs           |
-| `knownBugs`      | optional |  Bugs known of the test (e.g.: doesn't work in IE6)  |                                                    |
-| `notes`          | optional |                  Links to resources                  |                                                    |
+|                  | Necesity |                      Description                     |                           Notes                           |
+|------------------|:--------:|:----------------------------------------------------:|:---------------------------------------------------------:|
+| `name`           | required |             Name of the feature detection            |                                                           |
+| `property`       | required | The property name established in `Modernizr.addTest` |     Preferably just lowercase, without any punctuation    |
+| `tags`           | optional |    A group that encapsulates many feature detects    |                                                           |
+| `caniuse`        | optional |      A conversion table of caniuse and Modernizr     | Consider adding it to test/browser/integration/caniuse.js |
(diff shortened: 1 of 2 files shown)


metadata.md
```diff
@@ -0,0 +1,48 @@
...
+|                  | Necesity |                      Description                     |                           Notes                           |
+|------------------|:--------:|:----------------------------------------------------:|:---------------------------------------------------------:|
+| `name`           | required |             Name of the feature detection            |                                                           |
+| `property`       | required | The property name established in `Modernizr.addTest` |     Preferably just lowercase, without any punctuation    |
+| `tags`           | optional |    A group that encapsulates many feature detects    |                                                           |
+| `caniuse`        | optional |      A conversion table of caniuse and Modernizr     | Consider adding it to test/browser/integration/caniuse.js |
+| `authors`        | optional |          List of contributors of the script          |          There are a couple of ways to express it         |
+| `builderAliases` | optional |                                                      |                                                           |
```
MarkelFe, line 42: No idea what builder aliases are for

patrickkettner: they are aliases for the builder. [its for when tests are renamed](https://github.com/Modernizr/Modernizr/commit/cdb6ee49e255bc6aa278b9a406a975201a040715)

metadata.md
```diff
@@ -0,0 +1,48 @@
...
+| `polyfills`      | optional |     Available polyfills for not working versions     |               Add them in lib/polyfills.json              |
+| `aliases`        | optional |         Other names that the feature goes by         |                                                           |
```
MarkelFe, line 44: I'm not completely sure what are aliases used for, I think that their main function is to show them easier in the search, but maybe they have some kind of functionality that you may call the function by either of the 2 names. Just wanted to make sure.

patrickkettner: they should not be used for anything new. its a last resort for if something has multiple terms or things that the author thought people would use instead of the canonical name

MarkelFe: Maybe adding a _recommended_ or _not recommended_ option in the necessity column would be nice?

metadata.md
```diff
@@ -0,0 +1,48 @@
...
+/* DOC
+Here it would go a description of the feature test. You can use **markdown** here :)
+*/
+```
+> There is no set order in which they must appear, but it is common to see `name` and `property` at the top while `notes` at the bottom.
+
+## Item description
+|                  | Necesity |                      Description                     |                                  Notes                                  |
+|------------------|:--------:|:----------------------------------------------------:|:-----------------------------------------------------------------------:|
+| `name`           | required |             Name of the feature detection            |                                                                         |
```
patrickkettner, line 37: [technically this is not required](https://github.com/Modernizr/Modernizr/blob/870cf370ab370223199462897abfbf0ec5469878/lib/metadata.js#L69-L71). 

nothing really is, honestly.

metadata.md
```diff
@@ -0,0 +1,48 @@
...
+| `property`       | required | The property name established in `Modernizr.addTest` |            Preferably just lowercase, without any punctuation           |
+| `tags`           | optional |    A group that encapsulates many feature detects    |                                                                         |
+| `caniuse`        | optional |      A conversion table of caniuse and Modernizr     | Consider adding it to [caniuse.js](test/browser/integration/caniuse.js) |
+| `authors`        | optional |          List of contributors of the script          |                 There are a couple of ways to express it                |
```
patrickkettner, line 41: Im not sure I understand the purpose of the note.  Could you explain?

metadata.md
```diff
@@ -0,0 +1,48 @@
...
+| `builderAliases` | optional |     Used by CI and the web when tests are renamed    |                    Should not be needed in new tests                    |
+| `polyfills`      | optional |     Available polyfills for not working versions     |             Add them in [polyfills.json](lib/polyfills.json)            |
```
patrickkettner, line 43: i'd rephrase the note to something like "any polyfill listed needs to be included in polyfill.json"

metadata.md
```diff
@@ -0,0 +1,48 @@
...
+| `aliases`        | optional |   Used if a feature has more than a canonical name   |                    Should not be needed in new tests                    |
```
patrickkettner, line 44: note should be something like "legacy only - do not use"

metadata.md
patrickkettner: @rejas do you think metadata.js should throw if these are not provided?

metadata.md
```diff
@@ -0,0 +1,48 @@
...
+Here it would go a description of the feature test. You can use **markdown** here :)
+*/
+```
+> There is no set order in which they must appear, but it is common to see `name` and `property` at the top while `notes` at the bottom.
+
+## Item description
+|                  | Necesity |                      Description                     |                                  Notes                                  |
+|------------------|:--------:|:----------------------------------------------------:|:-----------------------------------------------------------------------:|
+| `name`           | required |             Name of the feature detection            |                                                                         |
+| `property`       | required | The property name established in `Modernizr.addTest` |            Preferably just lowercase, without any punctuation           |
```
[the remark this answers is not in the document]
MarkelFe: Okay, in that case you may want to updated [https://modernizr.com/docs](https://modernizr.com/docs) in the Modernizr API --> Add test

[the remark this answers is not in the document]
patrickkettner: that is not this.

metadata.md
MarkelFe: I mean, I used [this file](https://github.com/Modernizr/Modernizr/blob/870cf370ab370223199462897abfbf0ec5469878/test/node/lib/metadata.js#L159) to determine which things are necessary. Although it may be completely unrelated, I don't know the entire project architecture yet.

metadata.md
MarkelFe: That you may write it as `"Markel Ferro (@MarkelFe)"`, `"@rejas"` or `"Brandom Aaron"`. There isn't a single way to write it.

metadata.md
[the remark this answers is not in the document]
MarkelFe: Okay 😅

metadata.md
patrickkettner: the strings are never parsed, so I don't think you need to write that. The only thing I could think to mention is that is has to be a string, or array of strings

metadata.md
patrickkettner: good find! That tests the output of the metadata, this doc is about the input. What you linked to makes sure that what I linked to if performed. 
I am fine with it being explicitly required, but I think if we do we should make sure the existing metadata passes and make lib/metadata.js throw if it is missing

rejas: i think "required" should be the stuff that is absolutelty minium for the website to display something sensible. If it is missing then there should be a warning thrown

metadata.md
rejas: yeah, the author field is a free-for-all-clown fiesta sometimes :-) if there was ever a further usage of it besdies displaying it as a string then and only then it should be cleaned up.
for now I'd say though that we could politely ask for new submission to be of a certain format (like maybe "<name> (<github handle)")

rejas:
but why not put this info into the README or in a doc named "how to write tests" (working title) 
here in metadata.md it just sits around with no obvious purpose from the outside...

MarkelFe:
>  I didn't know where to put this information, so I created a new file,

I like more the idea of the "how to write tests" as I find that the README would get too big if we include the entire schema + table. Additionally, the information of this file is more technicalities than a general purpose read that a README represents.

But yes, even knowing of the existence of the file it was hard to find in the explorer, so a rename would be nice 😅.

patrickkettner:
> I find that the README would get too big 

+1.

rejas:
Then rename it to HOW_TO_WRITE_FEATURE_DETECTS.md :-)

MarkelFe:
Done, also I changed the header levels, because the title implies that in some day there will be other subjects apart from metadata

patrickkettner:
is the doc going to be updated to be about how to write feature detects, then? it doesn't really  mention that at all...

MarkelFe:
Okay, I'll add some title and description later in the day, writing Markdown in mobile is horrible 😅

patrickkettner:
thanks @MarkelFe!!

MarkelFe:
I made it a little vague because I'm not entirely sure which kind of information could go in this file, once it grows the statement could probably be more precise.

I've based the table contents style from the [atom contributing guidelines](https://github.com/atom/atom/edit/master/CONTRIBUTING.md). Also, sorry for the commit mess, I had just configured my GPG key on git and vscode and one commit went with the wrong email address.

MarkelFe:
@patrickkettner everything right with the PR?

rejas:
Last chance for @patrickkettner to object before I merge it tomorrow :-)

PR #2551, opened 2020-04-28
Require minimal values for metadata

### LINKED ISSUE #2548: Request a minimal amount of metadata
> i think "required" should be the stuff that is absolutely minimum for the website to display something sensible. If it is missing then there should be a warning thrown

_Originally posted by @rejas in https://github.com/Modernizr/Modernizr/pull/2547_

Opened an issue to have it separated from the PR. [Related code.](https://github.com/Modernizr/Modernizr/blob/870cf370ab370223199462897abfbf0ec5469878/lib/metadata.js#L69-L71)

At this moment the required values would be the name and property ones.

MarkelFe:
This PR closes #2548. 

Check the history if you want to see the original description.

Basically it requires the "name" and "property" values of the metadata to be filled. If they are null an error will be thrown.

The old system did `metadata.name = metadata.amdPath` if the name value was empty but `audio/ambientlight.js` is a horrible name for a feature detect.

The PR also edits the mocha:node tests in order to accommodate these changes (I've learnt that debugging tests is horrible)

### DIFF
--- a/lib/metadata.js
+++ b/lib/metadata.js
@@ -66,17 +66,10 @@ function metadata(cb) {
       metadata.path = './' + (start + '/' + file).replace(baseDir, '').replace(/\\/g, '/');
       metadata.amdPath = metadata.path.replace(/^\.\/feature\-detects/, 'test').replace(/\.js$/i, '');
 
-      // Force minimal metadata #2548
+      // Force minimal metadata #2551
+      // Consider adding new tests in test/node/lib/metadata.js if you add new minimal requirements
       if (!metadata.name || !metadata.property) {
-        process.emitWarning('\x1b[31mRequired values missing', {
-          code: 'MINIMAL_METADATA',
-          detail: file + ' is missing required values of metadata please refer to HOW_TO_WRITE_FEATURE_DETECTS.md. Check for typos too.\x1b[0m'
-        });
-      }
-
-      // Preserve to reduce number of errors
-      if (!metadata.name) {
-        metadata.name = metadata.amdPath;
+        throw new Error('Minimal metadata not found in `' + file + '`')
       }
 
       var pfs = [];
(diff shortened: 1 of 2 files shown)


MarkelFe:
Okay, Travis showed me something that did not happen in my PC, converting to draft while working on it

MarkelFe:
/ooo until tomorrow

But, first explain the error, okay in mocha:node it is tested a handful of metadata variants, which include versions without names and versions without properties, which is not great :( This results in a few 1200 lines of warning logs 😅

Solution:
- (Easy but bad) make the if that throws the warning not work when the mocha:node test is active, avoiding warnings. However that is not good code :/
- (Medium but meh) Locate the test script of mocha:node and make it send all metadata requests with a name and property value
- (Hard but the proper way) Adapt the mocha:node script to take the warning as a good result, however, by this time I would probably convert the warning in `throw new Error` because the polyfill value already does it, and maybe stopping the build for those typos isn't so bad

MarkelFe:
Okay I took the medium-hard path, I've made every test to give a `name` and `property` value, but I've created a pair of tests that check if these values are not sent. Everything should be working properly, no [1200 warnings will appear like before,](https://travis-ci.org/github/Modernizr/Modernizr/jobs/680704124#L268) and as far as my testing goes Modernizr won't compile if a `name` or `property` value is missing.

Marking ready for review

MarkelFe:
As a note, maybe the tests should be renamed, I got stuck 15 minutes because the test that was throwing the error was one of a duplicate name and I was looking in the wrong one (or maybe comments like I did for some of the tests).

I know that the errors give you the line of code of the test, but I think that it would he more friendly :)

patrickkettner:
@MarkelFe what test has a duplicate name?

MarkelFe:
`Should throw if we can't define` appears a pair of times (now it does 4, my fault 😅). I'll change the names tomorrow, the name, property and polyfill ones should be essay to rename

MarkelFe:
Gave a little more information about each test :)
Removed my previous comments as it doesn't make sense to maintain them 🙃

PR #2552, opened 2020-05-01
Metadata updates

MarkelFe:
(Yes, you don't have to mention it, I know that I'm maybe a little bit obsessed with metadata)

The PR modifies metadata, the original intention was to convert `warnings` that were `knownBugs` but I guessed I ended up looking at issues from 2012 and saying that ECMA 6 & 7 are no longer a draft (since 2015 % 16 🙄)

### DIFF
--- a/feature-detects/css/cubicbezierrange.js
+++ b/feature-detects/css/cubicbezierrange.js
@@ -5,7 +5,7 @@
   "tags": ["css"],
   "builderAliases": ["css_cubicbezierrange"],
   "authors": ["@calvein"],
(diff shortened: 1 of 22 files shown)


rejas:
I like to clean stuff up too, so no word from me against this PR :-)

MarkelFe:
I don't know how patrick did it, but I took a knownBug as something that goes wrong with the test itself (giving a false positive in specific version) while warning everything else, for example, that the browser doesn't have x functionality.

patrickkettner:
@MarkelFe is correct. a known bug is a known bad result, a warning is something like "this test is expensive to run, dont unless you need to"

PR #2560, opened 2020-05-12
Avoid testing in only docs files

MarkelFe:
This will avoid testing if only .md files (and another bunch) are edited. I know that [skip ci] exists but that makes the git log ugly.

Travis CI skip extracted from here: [https://github.com/google/EarlGrey/pull/383/files/3b38a5dea36a88aba42a42931e77a7c5429a1837](https://github.com/google/EarlGrey/pull/383/files/3b38a5dea36a88aba42a42931e77a7c5429a1837)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,14 @@ node_js:
 before_install:
   - npm install -g npm@latest
 before_script:
+  - |
+      if ! git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '(.md)|(.png)|(.ai)|(.svg)|(.jpg)|(.pdf)|^(LICENSE)|^(docs)|^(media)'
+      then
+        echo "Only doc files were updated, not running the CI."
+        exit
+      fi
   - npm install -g gulp-cli
 after_success:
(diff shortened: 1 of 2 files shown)


rejas:
Just wondering if we have any feature-detects/tests that rely on a file with one of these file extension...

MarkelFe:
I mean I highly doubt that there is, haven't found anything yet. Just to be sure we could remove the image file extensions as the media folder is already included.

Moreover, it's not that the files wouldn't be transfered to the CI is simply not to run the tests if, for example, someone only changes the README file. If someone edited a js file and a md file the build test would trigger.

rejas:
I'm still not convinced... I think only the md files should be included as to encourage them updating a frequently. But even then the tests dont cost much... what do you think @patrickkettner  ?

MarkelFe:
Okay, I'll modify it to only include MD files. Should the License count as a .md file too?

And yes, tests aren't very expensive but sometimes they fail for random things (for example, node 14 in travis is still a little buggy) as seen in my documentation PRs

MarkelFe:
> Okay, I'll modify it to only include MD files. 

Done, I haven't include the LICENSE file, just tell me and I'll reintroduce it

PR #2561, opened 2020-05-13
Fill the documentation

MarkelFe:
Adds these new sections to the HOW_TO_WRITE_FEATURE_DETECTS.md:

- Polyfills
- Submitting a PR (simple)
- Testing
- Install the basics (simplistic)
- Multiple rephrases for better readability

It also reorders some files to follow alphabetical order.


MarkelFe:
I'll soon solve the conflicts

MarkelFe:
Okay merge conflicts solved, still there is a broken link link with the `Verify/Run your tests` (seems that Github Markdown parser doesn't like slashes but the url redirect yes, etc) so I'll probably will change the name in the next commit (I assume that some extra tweaking will be needed).
Also, the install the basics section doesn't end up convincing me :/

HOW_TO_WRITE_FEATURE_DETECTS.md
```diff
@@ -36,20 +53,114 @@ The following code represents an example of the schema (it does not represent a
...
+},
+```
+
+> Metadata does no need to appear in a set order, however, it is common to follow the order stated above.
+
+### Polyfill Item Description
+
+|            | Necessity |                       Description                      |
+|------------|:---------:|:------------------------------------------------------:|
+| `name`     |  required | Name of the polyfill used in the [metadata](#metadata) |
```
MarkelFe, line 100: Okay, this is completely wrong

rejas: in what way?

MarkelFe: The name that you put in the name attribute is the one that appears on the web (if I'm not wrong), the name that it is used to name polyfills in the metadata is the head of the branch.
In this example you name the polyfill with `css3pie` not with `CSS3 PIE`, the latter one would be the prettified way that appears on the web.

MarkelFe: Solved in 6f838e7

MarkelFe:
🤔 I was thinking about the testing section, maybe it would be a good idea to put the `General testing` & `Caniuse testing` at level 4 under a `Create Tests` section (level 3) so it is more obvious that those 2 are for creating and different from the `verify and running` section.

Moreover, the testing section would be duplicated if we merge this PR [(check the README)](https://github.com/Modernizr/Modernizr#testing-modernizr)

rejas:
Since @patrickkettner went dark again, I'd merge it even without his review. Or what do you say @MarkelFe ?

MarkelFe:
Up to you @rejas 

I mean, I think that nothing is technically wrong, so suggestions and improvements (some even mentioned by me) can be added later down the road. But yeah, up to you 🙃

PR #2573, opened 2020-06-12
Add Clipboard API

### LINKED ISSUE #2572: Clipboard API
I was thinking that it would be interesting to create detections for the various Clipboard APIs.

Here are [the caniuse compatibility tables](https://caniuse.com/#search=clipboard%20API) and the [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API)

I would need to give a thought about how this could be approach, unfortunately, I'm going to be pretty busy until mid-July (time to catch up for all the time lost in quarantine) so I leave an issue so I don't forget + if someone is interested to give it a go.

Markel:
Closes #2572 

The PR introduces Clipboard API detection and it's sub-modules. It does it in an array like `inputtypes` does. 

### TODO
- [x] [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API)
- [x] [Read](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read)
- [x] [ReadText](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText)
- [x] [Write](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write)
- [x] [WriteText](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText)
~~And maybe... [Clipboard Event](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent), but I need to check it out first (probably yes)~~


feature-detects/clipboard/clipboard.js
```diff
@@ -0,0 +1,74 @@
...
+      // The sub-modules checked only if API is available to avoid Edge crashes
+      for (var i = 0; i < props.length; i++) {
+        if (navigator.clipboard[props[i]]) {
+            result = true;
+        } else {
+            result = false;
+        }
+        name = props[i].toLowerCase();
+        Modernizr.addTest('clipboard.' + name, result);
+      }
```
MarkelFe, line 38: If someone could help me with this part would appreciate it. As far as my debugging goes the if statement checks correctly whether the browser supports the feature, the problem seems to come more during the process of adding it to the array.

MarkelFe: Btw, to test it you'll need to uncomment lines 24 and 25 and probably comment the work around

rejas:
hi @MarkelFe sorry I am a little busy right now, lots of stuff to do at home.... Did you look at the discussion in https://github.com/Modernizr/Modernizr/pull/659 yet?

MarkelFe:
Don't worry @rejas I also have a incredibly busy month. And no, I hadn't looked at #659, it would have probably been a good idea to search before starting to work 😅. I'll take a look at it at some moment (probably not today)

feature-detects/clipboard/clipboard.js
MarkelFe: Okay, I've found the error 😅, i just had to remove Modernizr from line 37 🤦‍♂️. Submitting a PR soon.

MarkelFe:
Okay so the Clipboard API is completed. Here is an example of the test in Firefox 78 (without any flags changed):
Firefox 78 clipboard API detection

I'll probably take another look at #659 to implement the Clipboard Events detection so we can close 2 PR in one! (And also 2 feature detects in one) But the clipboard.js file should be finished review it without fear (if you have time, that nowadays is a difficult thing)

MarkelFe:
Okay, I've moved the feature detect to the root of feature-detects as I'm not implementing the Clipboard Event API, as the #659 shows how complicated it can be. If I decide to go for it I will create a new PR. So the PR is ready :)

MarkelFe:
Btw, don't know why the test fails, but seems a mocha error, specially as the last 172 all pass 🤷‍♂️

Markel:
Seems like the PR is getting staled, the PR is ready to go 🙂

PR #2575, opened 2020-06-22
Add detection for support of scrollToOptions mode of window.scrollTo()

### LINKED ISSUE #2433: Proposal: scrollToOptions detection

chris13524:
Fixes #2433
Replaces #2438

Thanks to @asmarcz for the main implementation!

### DIFF
--- a/feature-detects/scrolltooptions.js
+++ b/feature-detects/scrolltooptions.js
@@ -1,0 +1,34 @@
+/*!
+{
+  "name": "scrollToOptions dictionary",
+  "property": "scrolltooptions",
+  "caniuse": "mdn-api_scrolltooptions",
+  "notes": [{
+    "name": "MDN docs",
+    "href": "https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollTo"
+  }],
+  "authors": ["Oliver Tušla (@asmarcz)", "Chris Smith (@chris13524)"]
+}
+!*/
+define(['Modernizr', 'createElement', 'getBody'], function (Modernizr, createElement, getBody) {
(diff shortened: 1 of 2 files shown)


feature-detects/scrolltooptions.js
```diff
@@ -0,0 +1,34 @@
+/*!
+{
+  "name": "scrollToOptions dictionary",
+  "property": "scrolltooptions",
+  "caniuse": "mdn-api_scrolltooptions",
```
MarkelFe, line 5: I wonder if adding the mdn caniuse calls is of any utility 🤔. I mean, don't delete, it's okay, great job!
What I mean is that if we are not able to add them to the [main caniuse file](https://github.com/Modernizr/Modernizr/blob/master/test/browser/integration/caniuse.js) because they cause errors, what is the point adding them to the files?

rejas: The caniuse file is there for testing, the tag here is for displaying it on the website. So having a mdn here is ok in my opinion.

chris13524:
Can this be merged?

Markel (approved):
Oh sorry I distracted the PR with my comment and it never got merged 😓. Really sorry, its a nice feature Chris 🥳

rejas:
Merging it for v4

PR #2576, opened 2020-06-22
Solve some polyfill link issues

MarkelFe:
As many of the links go back years and years some websites are down or they updated their URLs, the PR changes the URLs or deletes the polyfills to accommodate the changes. Also it updates some links from http to https. It is important to notice that most github links haven't been checked because my link check extension was always hitting the GitHub API rate limit 🙃.

### DIFF
--- a/lib/polyfills.json
+++ b/lib/polyfills.json
@@ -68,7 +68,7 @@
   "css-escape": {
     "name": "CSS.escape()",
     "authors": ["Mathias Bynens"],
-    "href": "http://mths.be/cssescape",
+    "href": "https://github.com/mathiasbynens/CSS.escape",
     "licenses": ["MIT"],
     "notes": []
   },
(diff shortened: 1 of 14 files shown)


lib/polyfills.json
```diff
@@ -20,27 +20,10 @@
     "licenses": ["Apache2"],
     "notes": ["A small library that unifies the HTML5 Notifications APIs across different browsers including IE9 & IE10."]
   },
-  "css3pie": {
-    "name": "CSS3 PIE",
-    "authors": ["Jason Johnston"],
-    "href": "http://css3pie.com/",
```
rejas, line 26: that site still exists, why remove it?

MarkelFe: I haven't checked without deactivating my DNS blocklists but
[https://imgur.com/a/Rkt7Itz](https://imgur.com/a/Rkt7Itz)

MarkelFe: Okay, I've checked on my computer again and it works but it previously didn't? The Internet, full of mysteries... I'll re-add it

PR #2593, opened 2020-08-25
Support display:none iframes in flexgap test

### LINKED ISSUE #2590: Flex gap detection does not work when inside display:none iframe

chris13524:
Fixes #2590

### DIFF
--- a/feature-detects/css/flexgap.js
+++ b/feature-detects/css/flexgap.js
@@ -12,22 +12,35 @@
 }
 !*/
 define(['Modernizr', 'createElement', 'docElement'], function(Modernizr, createElement, docElement) {
-  Modernizr.addTest('flexgap', function() {
-    // create flex container with row-gap set
-    var flex = createElement('div');
-    flex.style.display = 'flex';
-    flex.style.flexDirection = 'column';
-    flex.style.rowGap = '1px';
+  Modernizr.addAsyncTest(function() {
+    function doWork() {
+      Modernizr.addTest('flexgap', function() {
+        // create flex container with row-gap set
+        var flex = createElement('div');
+        flex.style.display = 'flex';
+        flex.style.flexDirection = 'column';
+        flex.style.rowGap = '1px';
 
-    // create two elements inside it
-    flex.appendChild(createElement('div'));
-    flex.appendChild(createElement('div'));
+        // create two elements inside it
+        flex.appendChild(createElement('div'));
+        flex.appendChild(createElement('div'));
 
-    // append to the DOM (needed to obtain scrollHeight)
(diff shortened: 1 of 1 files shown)


Markel:
Lgtm 😄 as long as making it async doesn't make it backwards incompatible as we discussed in #2590 (which I doubt it but better to be sure than sorry).

The only thing is that inside the `else` comment (line 43) it would be nice to explain that the test doesn't work if we are inside a iframe:none and that is why we wait. Simply adding a `#2590` will probably be enough for people in the future to find out the reasoning.

chris13524:
@Markel good idea, I just pushed a comment linking to the issue

chris13524:
Re: backwards compatibility. I think the only way this would break compatibility is if they use `Modernizr.flexgap` during page load. Rather than returning `false` it would be `undefined`.

To potentially avoid this, we could adapt the test to be sync under normal circumstances and only be async if it detects it is inside a `display:none` iframe. However, this would still have the issue of `Modernizr.flexgap` not working under the async case. Also not sure if it is acceptable to have the async/sync nature switch out.

Or if there is some way to remove my `setTimeout()` call, that would be ideal as I think it would support using `Modernizr.flexgap` during page load, even though it is technically an async test.

Markel:
I don't really know how to contribute in here, I don't really know. Maybe @rejas knows better :)

Markel:
Maybe we should get back to this? 👀

rejas:
definitly. maybe @patrickkettner can take a look to :-) ?

Issue #2589, opened 2020-08-10
Cookie test doesn't currently set the cookie with required attributes for Chrome


dibsyjr1:
Google plan on blocking cookies in the near future that aren't set with certain attributes (SameSite = 'None' and Secure), the Cookie test doesn't currently set these properties showing warnings in dev tools that it'll be blocked in a future version of Chrome, could these attributes please be applied so that there's no problems in the near future with people's existing functionality.

Markel:
Yeah, we probably should solve this. Do you have a link to the announcement in which they detail the changes?

dibsyjr1:
I've got a couple of websites, there's the Chromium site where they detail the launch timeline as well as a bunch of helpful links about it here:
https://www.chromium.org/updates/same-site

Then there's also this Chromium blog post about preparing for the changes:
https://blog.chromium.org/2019/10/developers-get-ready-for-new.html

Markel:
I'm thinking that maybe we don't need to set up the cookie to be cross-site. The cookie should not be used in other websites (moreover, it is deleted once cookies are tested) and the need for a secure environment may result in bugs on HTTP sites.
Regardless, we should set up a `SameSite` value, as it is a good security practice, even if it is `strict`.

What do you think @dibsyjr1 ??

dibsyjr1:
Well the one issue I have is that i'm using the cookies test to make sure users have cookies enabled, otherwise I force redirect them to an error page. I admit asking for it to allow for cross-site might be a bit much, only other thing I can think of is if there were an option for either strict or cross-site? If not I can just work out some other solution, it may be that i'm the only one having this issue.

Markel:
I mean if they have cookies completely disabled `strict` cookies won't work. Maybe we can have options, although I admit that it may be impossible to check cross-site cookies, as we would need 2 domains (probably, not completely sure)

dibsyjr1:
Maybe it'd make more sense to just stick with strict, this seems more like a very specific issue that i'm having than a more widespread one.

Markel:
> only other thing I can think of is if there were an option for either strict or cross-site?

Hey, I just realized that we have a issue about that #2444, although I don't really know how that could be implemented without various servers

PR #2596, opened 2020-09-03
Set SameSite in cookie test

Markel:
Solves #2589 , kinda, but yes (feel free not to close the issue if you want)

The PR basically does what the title says, it really shouldn't affect as the cookie it's deleted once the test is performed, but hey the [Chromium team recommends it](https://blog.chromium.org/2019/10/developers-get-ready-for-new.html)

> However we strongly recommend you apply an appropriate SameSite value (Lax or Strict) and not rely on default browser behavior since not all browsers protect same-site cookies by default.

Moreover, the [caniuse page shows us](https://caniuse.com/#feat=same-site-cookie-attribute) that this feature is backwards compatible so it shouldn't result on any errors in old browsers

> This feature is backwards compatible. Browsers not supporting this feature will simply use the cookie as a regular cookie. There is no need to deliver different cookies to clients.

And if you are asking why I used `Lax` instead of `Strict` is because I prefer to be safe than sorry even if it shouldn't make any difference 😅

### DIFF
--- a/feature-detects/cookies.js
+++ b/feature-detects/cookies.js
@@ -24,10 +24,10 @@ define(['Modernizr'], function(Modernizr) {
     // or in sandboxed iframes (depending on flags/context)
     try {
       // Create cookie
-      document.cookie = 'cookietest=1';
+      document.cookie = 'cookietest=1; SameSite=Lax';
       var ret = document.cookie.indexOf('cookietest=') !== -1;
       // Delete cookie
-      document.cookie = 'cookietest=1; expires=Thu, 01-Jan-1970 00:00:01 GMT';
+      document.cookie = 'cookietest=1; SameSite=Lax; expires=Thu, 01-Jan-1970 00:00:01 GMT';
       return ret;
     }
     catch (e) {


dev-rke:
Hi,

i stumbled over your pull request as i have the same issue.
In my case i use a cross domain parent page for an iframe, where your implementation might break as well.
It's an assumption, i did not run your code, just set a cookie using chrome web developer tools within the iframe context.

To fix my case, i have to use SameSite=None and set the Secure flag:
```javascript 
document.cookie = 'cookietest=1; SameSite=None; Secure';
```

I assume your pull request will only fix cases where the iframe parent has the same origin.

Markel:
Hi @dev-rke may you elaborate on that? Because the cookie we set it should only be used for Modernizr purposes (actually it gets deleted after the test), moreover, setting the Secure flag would break compatibility with non-HTTPS websites.

dev-rke:
Hi @Markel ,

i did not use the whole pull request itself, but i used the main parts of the code (excluding Modernizr plugin setup code).
So i cannot assure if your implementation will work, as i did not test it.
My setup uses a cross domain iframe, so top domain and iframe domain are different. Both domains use HTTPS.
The cookie is set within the iframe. Using Chromium 86 on Ubuntu.

Setting document.cookie manually by using your code snippet with SameSite=Lax won't work within my setup, i had to use SameSite=None and the Secure flag to get it working.

dev-rke:
Just as update: i disabled the modernizr cookie test in my specific cross domain case, as it turns out that it seem not to work reliable, it might have other causes (maybe caching).
I'd recommend not to worry too much about SameSite=Lax vs None at the moment, both approaches will be an improvement. :-)

PR #2221, opened 2017-06-26
add try catch for datachannel test

philippsimon:
I got the following exception while running this test in the HbbTV browser of the Samsung 2015 und 2016 SmartTVs:
`NotSupportedError: DOM Exception 9`
Adding a `try catch` block solved that issue.

PR #2599, opened 2020-09-04
Fix datachannel test breaking Modernizr on PS4

guilhermesimoes:
On PS4 the following happens:

> new webkitRTCPeerConnection(null)
> TypeError: RTCPeerConnection argument must be a valid Dictionary
>
> new webkitRTCPeerConnection({})
> TypeError: Error creating RTCPeerConnection

Screenshot:
Screenshot 2020-09-04 at 15 06 59

Duplicate of https://github.com/Modernizr/Modernizr/pull/2221, but that PR has been open for 2 years and has conflicts.

### DIFF
--- a/feature-detects/webrtc/datachannel.js
+++ b/feature-detects/webrtc/datachannel.js
@@ -19,8 +19,12 @@ define(['Modernizr', 'prefixed', 'domPrefixesAll', 'test/webrtc/peerconnection']
     for (var i = 0, len = domPrefixesAll.length; i < len; i++) {
       var PeerConnectionConstructor = window[domPrefixesAll[i] + 'RTCPeerConnection'];
       if (PeerConnectionConstructor) {
-        var peerConnection = new PeerConnectionConstructor(null);
-        return 'createDataChannel' in peerConnection;
+        // Wrapped in a try catch to avoid "Error creating RTCPeerConnection" #2599 & #2221
+        try {
+          var peerConnection = new PeerConnectionConstructor({});
+          return 'createDataChannel' in peerConnection;
+        } catch (e) {
+        }
       }
     }
     return false;


feature-detects/webrtc/datachannel.js
```diff
@@ -19,8 +19,12 @@ define(['Modernizr', 'prefixed', 'domPrefixesAll', 'test/webrtc/peerconnection']
     for (var i = 0, len = domPrefixesAll.length; i < len; i++) {
       var PeerConnectionConstructor = window[domPrefixesAll[i] + 'RTCPeerConnection'];
       if (PeerConnectionConstructor) {
-        var peerConnection = new PeerConnectionConstructor(null);
-        return 'createDataChannel' in peerConnection;
+        // PS4 throws "Error creating RTCPeerConnection" when attempting to create an instance of webkitRTCPeerConnection
```
Markel, line 22: Seems to be more than the PS4 error, maybe a better comment could be `Wrapped in a try catch to avoid "Error creating RTCPeerConnection" #2599 & #2221`. What do you think? 🤔😊

guilhermesimoes: That's better!

feature-detects/webrtc/datachannel.js
```diff
@@ -19,8 +19,12 @@ define(['Modernizr', 'prefixed', 'domPrefixesAll', 'test/webrtc/peerconnection']
     for (var i = 0, len = domPrefixesAll.length; i < len; i++) {
       var PeerConnectionConstructor = window[domPrefixesAll[i] + 'RTCPeerConnection'];
       if (PeerConnectionConstructor) {
-        var peerConnection = new PeerConnectionConstructor(null);
-        return 'createDataChannel' in peerConnection;
+        // Wrapped in a try catch to avoid "Error creating RTCPeerConnection" #2599 & #2221
+        try {
+          var peerConnection = new PeerConnectionConstructor({});
+          return 'createDataChannel' in peerConnection;
+        } catch (e) {
```
rejas, line 26: if it errors shouldnt it return false then since the test fails?

guilhermesimoes: No, because this is inside a loop. We don't want to return `false` if the first iteration fails, we want to try them all.

rejas:
So, is this something that you need soonish or can it wait for an unknown amount of time until v4 is released?

Markel:
> So, is this something that you need soonish or can it wait for an unknown amount of time until v4 is released?

As of some hope I already implemented the changes to the v4 PR 346f0c7. But yeah, the v4 version release date is pretty much unknown

PR #2600, opened 2020-09-04
Remove old test for String.prototype.contains

guilhermesimoes:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/contains
redirects to
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

Also see https://github.com/Modernizr/Modernizr/pull/2136


Markel:
So, which is the reason to remove it?

Btw, to remove it you need to delete it from `lib/config-all.json` and check whether the polyfill is used elsewhere and if not remove it from the config.

guilhermesimoes:
> So, which is the reason to remove it?

["Includes" was originally going to be named "contains". Unfortunately (...) Array.prototype.contains was used by Mootools in a way that caused this to be a breaking change. If you don't remember Mootools, at one point it was a pretty popular framework. It's not heavily used now, but a ton of sites on the web were built with it and have used it for years without active maintenance. It is abandoned but valuable. This would break them overnight. The standards committee decided to prioritize a functional web over the perfect syntax, and instead changed "contains" to "includes" to preserve functionality that everyone expected to just keep working.](https://www.telerik.com/blogs/web-going-2016#toc_4)

Markel:
Makes sense to me :)

README.md
```diff
@@ -23,6 +23,7 @@ Modernizr tests which native CSS3 and HTML5 features are available in the curren
     - `touchevents`: [discussion](https://github.com/Modernizr/Modernizr/pull/2432) 
     - `unicode`: [discussion](https://github.com/Modernizr/Modernizr/issues/2468) 
     - `templatestrings`: duplicate of the es6 detect `stringtemplate`
+    - `contains`: duplicate of the es6 detect `es6string`
```
rejas, line 26: should be "duplicate of es6string.include"

Issue #2595, opened 2020-09-02
Add simple test for the property value


Markel:
Hi 👋

> Note: Check the comment below

### What is proposed
This should be a relatively easy PR to get done. Every Modernizr test has a `property` tag on their [metadata](https://github.com/Modernizr/Modernizr/blob/master/HOW_TO_WRITE_FEATURE_DETECTS.md#metadata), this tag is always* lowercase without any punctuation. The idea would be to create a test that throws an error* whenever a property doesn't respect by this rule.

*[The Modernizr website says that this rule is just a suggestion](https://modernizr.com/docs#modernizr-addtest) but in reality it's more a rule than a recommendation. Feel free to throw a warning instead of an error if you prefer.

### How to do it
This test should be implemented in the [lib\metadata.js file](lib/metadata.js), a recent similar PR was approved (#2551) which added a similar test to this. After adding the test we would need a test for the test, this one would need to be implemented in [test\node\lib\metadata.js file](test\node\lib\metadata.js), you may check #2251 too (although the git diff is less readable in this one)

Feel free to raise any question in this issue if you stumble with problems :)

Markel:
Okay, I should have checked before writing an issue 🤦. [An error is actually thrown if you put uppercase symbols](https://github.com/Modernizr/Modernizr/blob/1baf805e9eb54710f6ad122a4f6557f6e369c530/test/browser/integration/classes.js#L12-L14), but not for symbols like spaces or others. So yeah, the issue should continue, but you may want to [check this issue](https://github.com/Modernizr/Modernizr/pull/2108#issuecomment-619512863) to do the work. (It may actually be easier than before!)

adzo261:
Hi, can I take this issue?

rejas:
Sure, go for it

rejas:
Merged!

Markel:
I don't know, but I quite like this format to create small changes that need to be done. Like explaining the things related to the change and give like some idea from where to start. What do you think @rejas ? @adzo261, maybe you could give us some feedback :)

adzo261:
> I don't know, but I quite like this format to create small changes that need to be done. Like explaining the things related to the change and give like some idea from where to start. What do you think @rejas ? @adzo261, maybe you could give us some feedback :)

[...]

PR #2601, opened 2020-09-06
Add valid chars test for the property of metadata

adzo261:
Refers to issue #2595

As per our discussion, `metadata.property` can contain only `lowercase alphanumeric characters and dashes`.
PR adds a test for the same, which raises an error if `metadata.property` values doesn't have valid characters. 

Please have a look. Thanks!

### DIFF
--- a/lib/metadata.js
+++ b/lib/metadata.js
@@ -84,11 +84,11 @@ function metadata(cb) {
         });
       }
 
-      //Raise error if metadata.property is NOT a lowercase string
+      //metadata.property can only have lowercase alphanumeric and dashes
       const properties = Array.isArray(metadata.property) ? metadata.property : [metadata.property];
       properties.forEach(function (property) {
-        if (property.match(/[A-Z]/)) {
-          throw new Error(metadata.name + ' : ' + property + ': Property must be a lowercase string ' )
+        if (!property.match(/^[a-z0-9-]+$/)) {
+          throw new Error(metadata.name + ' : ' + property + ': Property can only have lowercase alphanumeric characters and dashes ' )
         }
       })
 
--- a/lib/metadata.js
+++ b/lib/metadata.js
@@ -83,6 +83,7 @@ function metadata(cb) {
           }
         });
       }
+      metadata.polyfills = pfs;
 
       //metadata.property can only have lowercase alphanumeric and dashes
       const properties = Array.isArray(metadata.property) ? metadata.property : [metadata.property];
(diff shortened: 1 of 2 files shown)


lib/metadata.js
```diff
@@ -83,6 +83,15 @@ function metadata(cb) {
           }
         });
       }
+
+      //Raise error if metadata.property is NOT a lowercase string
+      const properties = Array.isArray(metadata.property) ? metadata.property : [metadata.property];
+      properties.forEach(function (property) {
+        if (property.match(/[A-Z]/)) {
```
Markel, line 90: 🤔 I was wondering why you used a `forEach` to analyze the properties. Each file is already checked separately thanks to line 11.  I think that we would only need current lines 90 to 92 to do the task in this case (+ in that way we have a [much cleaner output error](https://imgur.com/a/x0d5IbI))

Markel: Btw, just as we are here I've seen in the other file that in the test you included `_`, maybe we could throw spaces and `_` to the Regex (as we really don't use them either). Take into account that we would need to change the error message to something like `Property includes invalid characters`

Markel: But for the rest, really good work Aditya!! 💪. Thank you very much :)

adzo261: Thank you for reviewing and commenting.
I found that `metadata.property` was an array of strings in somecases.
And in other cases it was a normal string.
Hence, I had to add a check for array and run a loop over it.


And do we need to check symbols(non-alphanumeric) also? I can add that check too if we want it. And then we can change the error to more generic as you said.

Markel: > I found that `metadata.property` was an array of strings ...

Ah okay, I didn't think about that. Very clever! In that case that part is perfect.

> And do we need to check symbols(non-alphanumeric) also?

It would be nice to test for symbols too (except the dash `-` as we use kebab-case for properties). If you include dots take care, I don't think so, but maybe subproperties (like in the clipboard test) give you errors.

> Thank you for reviewing and commenting.

Thanks to you for working on the PR :)

adzo261: From what I could understand, `metadata.property` can have only lowercase alphanumeric characters and a special symbol `-` for kebab-case. <br> Can you once confirm the set of symbols we can't have? <br>
And also does - `"Property can have only lowercase alphanumeric characters and kebab symbol"` makes sense as an error message then?

Post confirmation, I will start working on it.

Markel: Yep, if I'm not mistaken only lowercase letter and dashes are allowed.

About the error message, maybe is more clear if it says `dashes` instead of `kebab symbols`.

adzo261: I have updated the PR. We can have a second round of review.

Markel (approved):
Looks okay to me, I specially like the you dealt with the regex there :) If I had to nitpick maybe I would move line 95 to 86 (to keep the polyfill test along). But for me this is an approval 🎉

What do you think @rejas ??

rejas:
Looks good. Nice work @adzo261

PR #2603, opened 2020-09-12
Implement issue templates

Markel:
As we discussed in #2595 I've worked in some issue templates (haven't done anything about PR templates)

I've included 3 templates:
- Bug reports (caniuse errors, false positives & general bugs)
- Test requests
- Good first issues (although it may be used as a template for answers)

Another template that could be implemented is general requests (like coverage implementations or others), but they are not very common so I tried to minimize the decision problem for new contributors. Also, in the bug reports, maybe a Github Actions or something could be use to auto-assign labels to the issues (as multiple labels could be use for that template)


Markel:
OH! A questions template would be perfect!

PR #2607, opened 2020-10-05
Add feature detection for IndexedDB 2.0

### LINKED ISSUE #2571: add support for indexeddb 2.0
https://caniuse.com/#search=indexeddb

Now the indexeddb detect script only detects basic indexeddb apis.

But some browser versions implement 1.0 and some 2.0. 

So we can add a new script to detect 2.0 support.

In my opinion, 2.0 add some new apis compared to 1.0. Such as `getAll/getAllKeys`.

So we can use `'getAll' in IDBIndex.prototype` to determine whether the browser supports it.

Xenonym:
Closes #2571.

### DIFF
--- a/feature-detects/storage/indexeddb2.js
+++ b/feature-detects/storage/indexeddb2.js
@@ -1,0 +1,22 @@
+/*!
+{
+  "name": "IndexedDB 2.0",
+  "property": "indexeddb2",
+  "tags": ["storage"],
+  "caniuse": "indexeddb2",
+  "authors": ["Tan Zhen Yong (@Xenonym)"],
+  "polyfills": ["indexeddb"],
+  "async": true
+}
+!*/
+/* DOC
+Detects support for the IndexedDB 2.0 client-side storage API.
+*/
+define(['Modernizr', 'addTest', 'test/indexeddb'], function(Modernizr, addTest) {
+  Modernizr.addAsyncTest(function() {
+    Modernizr.on('indexeddb', function(result) {
+      if (!result) return;
+      addTest('indexeddb2', 'getAll' in IDBIndex.prototype);
+    });
+  });
+});
--- a/lib/config-all.json
+++ b/lib/config-all.json
@@ -272,6 +272,7 @@
     "serviceworker",
     "speech/speech-recognition",
     "speech/speech-synthesis",
+    "storage/indexeddb2",
     "storage/localstorage",
     "storage/sessionstorage",
     "storage/websqldatabase",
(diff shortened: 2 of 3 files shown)


feature-detects/indexeddb2.js
```diff
@@ -0,0 +1,22 @@
+/*!
+{
+  "name": "IndexedDB 2.0",
+  "property": "indexeddb2",
+  "tags": ["storage"],
+  "caniuse": "indexeddb2",
+  "authors": ["Tan Zhen Yong (@Xenonym)"],
+  "polyfills": ["indexeddb"],
```
rejas, line 8: is that polyfill also working for indexedDB 2.0?

feature-detects/indexeddb2.js
```diff
@@ -0,0 +1,22 @@
+/*!
+{
+  "name": "IndexedDB 2.0",
+  "property": "indexeddb2",
+  "tags": ["storage"],
+  "caniuse": "indexeddb2",
```
rejas, line 6: please add this also to the test/browser/integration/caniuse.js file

feature-detects/indexeddb2.js
Xenonym: Yep! IndexedDBShim has 2.0 only methods like [`getAll()` and `getAllKeys()`](https://github.com/indexeddbshim/IndexedDBShim/blob/master/src/IDBIndex.js#L429).

feature-detects/indexeddb2.js
Xenonym: It's already in the PR: https://github.com/Xenonym/Modernizr/blob/feature/add-indexeddb-2/test/browser/integration/caniuse.js#L99

rejas:
Looks good. One last thing though: Could you move the test into the storage subdirectory? seems to fit in there. Yes, indexeddb v1 isnt in there yet but we can clean that up later before v4

Xenonym:
> Could you move the test into the storage subdirectory? 

Moved!

Markel:
Nice to see some hacktoberfest contributions :)

rejas:
any comments from you @Markel about this PR? think it can be merged?

Markel:
Sorry for not getting back, kinda of a busy week with exams (writing in a train right now :)

I like the detect :) It has already been merged so congrats @Xenonym on your first commit  to the repository 🥳🥳 and thank you 😊

Btw @alsotang you may want to check this out :)

alsotang:
the detect method is right. thank you.

Markel F. <notifications@github.com>于2020年10月31日 周六04:07写道：

PR #2608, opened 2020-10-05
`indexeddbblob`: Fix test not triggering, update metadata

Xenonym:
* `indexeddbblob`: Fix test never triggering

	The `indexedbblob` test never triggers, even on platforms that support
	IndexedDB or blobs in IndexedDB databases.
	
	This is because in 164e757, `indexeddb` was changed to an async test,
	but `indexeddbblob` still assumes that `indexeddb` test is synchronous.
	Therefore, the `indexeddb` check is never true and `indexedbblob` never
	triggers.
	
	The `onsuccess` / `onerror` / exception handlers also assign the test
	result to a shared variable before calling `addTest()`, which results
	in the test result being always false as the variable will not be
	updated synchronously before `addTest()`.
	
	Let's:
	* wrap the `indexeddbblob` test in a `on('indexeddb.deletedatabase')`
	listener, so that it will trigger when `indexeddb` and
	`indexeddb.deletedatabase` pass
	* call `addTest()` in the `onsuccess` / `onerror` / exception handlers
	instead of assigning to a shared variable

* indexeddbblob: Add missing metadata properties

	* add `storage` tag
	* mark as `async`

### DIFF
--- a/feature-detects/indexeddbblob.js
+++ b/feature-detects/indexeddbblob.js
@@ -14,54 +14,52 @@ define(['Modernizr', 'addTest', 'prefixed', 'test/indexeddb'], function(Moderniz
   // For speed, we don't test the legacy (and beta-only) indexedDB
 
   Modernizr.addAsyncTest(function() {
-    var indexeddb;
-    var dbname = 'detect-blob-support';
-    var supportsBlob = false;
-    var openRequest;
-    var db;
-    var putRequest;
+    Modernizr.on('indexeddb.deletedatabase', function(result) {
+      if (!result) return;
 
-    try {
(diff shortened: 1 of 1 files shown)


rejas:
Thx @Xenonym  that sounds very reasonable. What do you think @patrickkettner sicne you wrote the original tests?

Xenonym:
@rejas would you be so kind as to label this `hacktoberfest-accepted` or approve it if this is not going to be merged before Oct 31?

Monib007:
hey could you assign me this task and let me contribute?

rejas:
> hey could you assign me this task and let me contribute?

could you look for another issue. This PR is stale and not currently worked but still somebody else worked on it.

PR #2610, opened 2020-10-14
Add detect/tests for prefetch

### LINKED ISSUE #2536: Rel Prefetch Support?
Any chance we could get a detect for rel prefetch? It'd be super-helpful for WebKit browsers. There's a [bug out for it](https://bugs.webkit.org/show_bug.cgi?id=194539) now, but no movement yet.

HeroProtagonist:
Closes #2536

- Made a detect for prefetch based on https://github.com/w3c/preload/issues/7
- Added general test and caniuse test


robin-drexler:
Hey, just stumbled upon this. This check may exclude IE11 which appears to [support prefetch](https://caniuse.com/link-rel-prefetch), but [not relList](https://caniuse.com/rellist) :( 

This isn’t an issue for preload because IE supports neither.

HeroProtagonist:
Do any of the detects do browser detection for a case like this?
I see a User Agent parser here, but not in source code

https://github.com/Modernizr/Modernizr/blob/94592f279a410436530c7c06acc42a6e90c20150/test/browser/integration/caniuse.js#L4

feature-detects/rel/prefetch.js
```diff
@@ -0,0 +1,27 @@
+/*!
```
rejas, line 1: creating a REl directory for this test only seems coutnerintuitive. isnt there a better palce for this?

rejas:
yeah, that UAparser is onyl used during testing of Modernizr to check against the caniuse database. It is generally not advised to use it in normal feature tests.

HeroProtagonist:
> yeah, that UAparser is onyl used during testing of Modernizr to check against the caniuse database. It is generally not advised to use it in normal feature tests.

Looks like `navigator.userAgent` is used in some detections. Maybe that is a better option

https://github.com/Modernizr/Modernizr/search?q=navigator.userAgent

feature-detects/rel/prefetch.js
HeroProtagonist: I can move it, just was thinking a follow up pr would be for `rel=preload`

rejas:
Yes, that sounds reasonable :-)

feature-detects/rel/prefetch.js
rejas: Then I propose putting it in a "link" subdirectory, since it is a test on the link element ...

HeroProtagonist:
@rejas Added ie11 check using `document.documentMode` which is used in other parts of the code for similar IE specific detection

feature-detects/rel/prefetch.js
HeroProtagonist: done!

rejas:
Thanks for the PR and your persistence :-) Merged!

Issue #2460, opened 2019-06-27
Incorrect Regular Expression in `/lib/build.js`


PatOnTheBack:
Line 42 and 43 of `/lib/build.js` says:

```JavScript
// remove last return statement and trailing })
contents = contents.replace(/return.*[^return]*$/, '');
```

[lgtm spotted a duplicate r](https://lgtm.com/projects/g/Modernizr/Modernizr/snapshot/8b121bd16ccc619bef5060e78eccb6420c4e7b87/files/lib/build.js?sort=name&dir=ASC&mode=heatmap) inside `[^return]`.

Should the lines be like this instead?

```JavaScript
// remove last return statement and trailing })
contents = contents.replace(/return.*([^r][^e][^t][^u][^r][^n])*$/, '');
```

rejas:
This proposed new line looks overly complicated. Why is the old one a problem?

PatOnTheBack:
@rejas The old line is a problem because 'r' is repeated inside of the negated list `[^return]*$`. With the way the regular expression is written, the final asterisk allows the negated set to be ignored completely.
The old regular expression matches:

- "return apple"
- "return return" // This should not be matched.
- "return alert"

I don't think my proposed regular expression fixed the problem, but the regex definitely needs to be fixed. One possible way to fix it is using [negative look-arounds](https://stackoverflow.com/questions/406230/regular-expression-to-match-a-line-that-doesnt-contain-a-word).

rejas:
Thx for the info. Would you be willing to help and to create a PR for this?

PatOnTheBack:
@Rejas I am willing to help with a PR. To know how to best write the regex, I need to understand why the matching needed to be performed in the first place. The code was not commented, so I wasn't entirely sure what the regex was for.

knittl:
Indeed, this regular expression looks like it was intended match any `'return'` not followed by another `'return'`. But `[^return]` does not mean "not the string 'return' at this position", but rather "not any of the characters in the group at this position".

It would be really helpful to have a list of expected input and output. As mentioned earlier, it is not exactly clear what the code is supposed to do; the assumption being "remove the everything from a line after the last 'return' in this line". Here's my attempt to define input=>output mappings:

* 'return me return not}' => 'return me'
* 'return 0;}' => ''
* 'return return;' => 'return'
* 'return {};}' => ''
* 'return apple' => ''
* 'return alert;}' => ''
* 'more return' => 'more'
* 'more return me' => 'more'
[...]

PR #2613, opened 2020-10-25
Fix regex to remove last return statement #2460

knittl:
Rewrite regex with character group used as negative lookahead. Instead,
match every line with return in it and then remove everything after the
last 'return' occurrence with a callback function.


lib/build.js
```diff
@@ -39,8 +39,12 @@ var baseRequireConfig = {
...
 
       if (!contents.match(/Modernizr\.add(Async)?Test\(/)) {
-      // remove last return statement and trailing })
-        contents = contents.replace(/return.*[^return]*$/, '');
+        // remove last return statement and trailing })
+        contents = contents.replace(
+          /^.*return.*$/,
+          function(match) {
+            return m.split("return").slice(0, -1).join('return').trim();
+          });
```
knittl, line 47: Another option would be implementing the following replace call:

```javascript
contents = contents.replace(/^(.*)return.*$/, '$1');
```

The only caveat is, that `'return yes return no'` becomes `'return yes '` with a trailing space, not sure if that would be a problem. Using the callback version is definitely the most flexible.

PR #2630, opened 2021-01-10
Fix `Undefined identifier` error in es6symbol in IE8

toddwong:
Keyword 'for' in `Symbol.for` caused `Undefined identifier` error in IE8.

https://github.com/Modernizr/Modernizr/blob/3514072f7ffab2b4e222c1be190c3bcb74ee83e5/feature-detects/es6/symbol.js#L23


Markel:
Thank you @toddwong !!

Did you recheck to see if it worked in IE8 (if you didn't just check it on the console, don't bother building the library). It looks okay, although I admit that I would never had thought of that solution.

toddwong:
Yes, I rechecked.  I actually rebuilt the library. We rebuild Modernizer everytime the website is packed with a webpack plugin.

rejas:
Thx for the PR, will merge it in the v3 branch too and release a bugfix version monday.

PR #2633, opened 2021-01-18
Fix a typo (probably)

Markel:
I think that the `es6class` is misspelled, I just noticed :)


Markel:
Also, a good test for #2560

PR #2639, opened 2021-01-30
Cleanup feature detects

rejas:
Rearranged some tests that belong in subdirectories IMHO. 

Any other candidates in your opinion @patrickkettner @Markel ?


lib/config-all.json
```diff
@@ -38,11 +38,10 @@
     "canvas/blending",
     "canvas/todataurl",
     "canvas/winding",
-    "canvastext",
+    "canvas/text",
```
Markel, line 41: emoji.js requieres canvas/text, that's the why to the errors

Markel (changes requested):
- Is there a reason why the audio.js test isn't in the audio folder?
- Same for the canvas.js, battery.js, video.js, input.js tests
I mean the main reason why I say it is because it happened to me searching for them + we are moving indexeddb to its folder

- Also inputsearchevent.js & inputtypes.js probably deserve to go in the input folder too

README.md
```diff
@@ -25,7 +25,13 @@ Modernizr tests which native CSS3 and HTML5 features are available in the curren
   
   - `class` to `es6class` to keep in line with the rest of the es-tests
 
-- These tests got removed:
+- Following tests got moved in subdirectories:
+
+  - `cookies`, `indexeddb`, `indexeddbblock`, `quota-management-api`, `userdata` moved into the storage subdirectory
```
Markel, line 30: Seems like autocorrect has autowrong in indexedblob 😅. Just noticed.

Markel (changes requested):

README.md
rejas: Good catch!

Markel (approved):
lgtm

rejas:
>     * Is there a reason why the audio.js test isn't in the audio folder?
> 
>     * Same for the canvas.js, battery.js, video.js, input.js tests
>       I mean the main reason why I say it is because it happened to me searching for them + we are moving indexeddb to its folder
> 
>     * Also inputsearchevent.js & inputtypes.js probably deserve to go in the input folder too

Moved some stuff. Just wondering with the input tests: Becuase we also have a forms subdirectory, so to me it looks pretty unclear which goes where :-) Maybe move all the feature detects (including the ones from forms) into input?

Markel:
> Maybe move all the feature detects (including the ones from forms) into input?

Yeah, we could, I mean I don't think that they cause the biggest confusion but what you say definitely makes sense.

rejas:
I think I got them all @Markel @patrickkettner 
Anything left to do in your opinion before I merge?

Markel (approved):
Looks good :)

PR #2646, opened 2021-04-10
Replace 'kebab-case' to 'hyphen-case'

DervishD:
Since 'kebab-case' may be perceived as derogatory by some members of
Arabic or Asian cultures, it's better to use a culture-agnostic term
like 'hyphen-case', which in addition is more self-explanatory than the
potentially offensive term 'kebab-case', which makes no sense without
knowing the origin of the term.


Markel (approved):
Glad to see that you managed to submit the PR (and yes, VSCode is very handy indeed).

Seems Unix vs. Windows line endings in changelog have tricked you a little bit (do not worry about it, it doesn't matter, I just comment it to have it searchable in the future).

The PR is great, glad to have contributions like this! 🎉

DervishD:
Markel, thanks a lot for your help :)

About line endings, I thought I had my local git configured as to not mess with line endings, using whatever the repository had, and if none, forcing Unix ones.

I have checked the local copy I cloned and in fact `Changelog.md` has UNIX line endings, so maybe something got broken when I committed or pushed, I don't know.

Sorry for the inconvenience.

rejas:
Thx for the PR @DervishD

DervishD:
My pleasure, @rejas :)

Issue #2656, opened 2021-07-09
Minification breaks config during generation


t3chguy:
Version 3.11.7

Related https://github.com/Modernizr/Modernizr/issues/1663 https://github.com/Modernizr/Modernizr/issues/2412

With the following config:

```json
{
  "minify": false,
  "enableClasses": false,
  "feature-detects": [
    "test/css/animations",
    "test/css/displaytable",
    "test/css/filters",
    "test/css/flexbox",
    "test/css/objectfit",

    "test/es5/date",
    "test/es5/function",
    "test/es5/object",
    "test/es5/undefined",

    "test/es6/array",
    "test/es6/collections",
    "test/es6/promises",
    "test/es6/string",

    "test/svg",
    "test/svg/asimg",
    "test/svg/filters",

    "test/url/parser",
    "test/url/urlsearchparams",

    "test/cors",
    "test/crypto",
    "test/iframe/sandbox",
    "test/json",
    "test/network/fetch",
    "test/storage/localstorage"
  ]
}
```

Generation without minification works correctly and has the following `_config`:

```
_config: {
  'classPrefix': '',
  'enableClasses': false,
  'enableJSClass': true,
  'usePrefixes': true
},
```

But if I enable minification it changes to:

```
_config:{classPrefix:"",enableClasses:!0,enableJSClass:!0,usePrefixes:!0}
```

Which has `enableClasses` set to `true` - this only ends up affecting rules added via `addTest` however so ends up with a weird split-brained behaviour.

PR #2657, opened 2021-07-09
Fix config generation when minifying

t3chguy:
Fixes https://github.com/Modernizr/Modernizr/issues/2656


rejas:
Thx for the PR, very much appreciated. Will pick it into the v3 branch and release a new version "soonish" (aka tomorrow if nothing goes wrong)

rejas:
By the way: Do you feel adventourous enough to write a test for this so the code coverage goes up again?

rejas:
Ok, "soonish" came quicker than I thought :-) Hope v3.11.8 helps your project a little!

t3chguy:
Hah thanks, I ended manually mutating the minified output, and given we only rebuild modernizr every six months or so it'll come in handy for when I would have inevitably forgotten to mutate the minified output in another six months

PR #2666, opened 2021-10-03
Changed To ES6

aalimaslam:


patrickkettner:
Thanks for the contribution! The point of modernize it is for the checks to work in older browsers￼. Is there a reason you're making this check in a single file location??￼

rejas:
Closing due to inactivity and no answer from the OP.

PR #2670, opened 2021-10-09
Use standard LICENSE text for MIT license

reviewher:
GitHub does not currently recognize the license text as MIT (it currently says "View license" instead of "MIT License")

gh-modernizr

The license parser is very persnickety.  This particular rendition is from [VueJS](https://github.com/vuejs/vue/blob/dev/LICENSE) which is an MIT-licensed project according to GitHub

This also addresses an issue where the package.json files field includes `LICENSE` but not `LICENSE.md`


rejas:
Thx for the PR!

PR #2671, opened 2021-10-10
Add a|area[ping] detection

### LINKED ISSUE #2484: <a ping=""> detection
Quick note here as I don't have time to write this up into a proper PR yet.

[Hyperlink auditing](https://html.spec.whatwg.org/multipage/links.html#hyperlink-auditing) allows for a parallel background request to be made on a hyperlink click, often used for tracking and click-detection. It is currently not supported by modernizr.

According to [this SO post](https://stackoverflow.com/questions/57608000/feature-detection-of-a-ping-hyperlink-auditing), the expression:

    "ping" in HTMLAnchorElement.prototype

will return `true` if the feature is supported, `false` otherwise. Empirically, we find this to be true.

heliocorreia:
Fixes #2484

### DIFF
--- a/feature-detects/a/aping.js
+++ b/feature-detects/a/aping.js
@@ -5,6 +5,7 @@
   "caniuse": "ping",
   "tags": ["media", "attribute"],
   "builderAliases": ["a_ping"],
+  "authors": ["Hélio Correia (@heliocorreia)"],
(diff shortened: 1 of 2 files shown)


heliocorreia:
@rejas done! ;)

PR #2674, opened 2021-10-12
Add new feature-detects from main branch

rejas:


rejas:
Weird, why do the tests fail so miserably? Any idea @Markel ?

PR #2677, opened 2021-10-16
Set gulp connect to listen on any IP addresses available

heliocorreia:
Set gulp connect to listen on any IP addresses available, i.e., the gh-pages and browser tests can be accessed from localhost and virtual machines when pointing to the host IP address.

https://www.npmjs.com/package/gulp-connect#optionshost


heliocorreia:
It is impossible to reach the gulp connect server instance when using a browser from a virtual machine (integrations tests) because the default host is localhost/127.0.0.1.

rejas:
Thx for the explanation! Merging it :-)

PR #2688, opened 2022-02-25
Fix: Update lowbandwidth condition to match new spec

jfkm69:
The connection spec has gone through mutiple updates but the lowbandwidth test still tests against the initial spec which introduced the preliminary version of the network information API. The newest version of the NetworkInformation API, which is still in draft (https://wicg.github.io/netinfo/#dfn-connection-type), has changed the name of the property being used to determine the quality of the network from `type` to `effectiveType`. Kindly note `type` now provides a list of strings describing different network adapters (ie ethernet, bluetooth, wifi, e.t.c). This PR updates the check to account for the old and new properties.


rejas:
Thx for the PR!

PR #2708, opened 2022-10-14
Adds PushManager feature detection #2703

### LINKED ISSUE #2703: Detect PushManager
**Which feature do you want to detect?**
It would be helpful to know if PushManager is available for notifications.

**Which browsers do support this feature?**
Chrome, Firefox should as well, but it has no "PushManager" in window.

**How could it be implemented?**
I'm not sure why Firefox has no "PushManager" even though canIuse shows browser support.

dawidkulpa:
This PR adds PushManager detection and resolves #2703.

MDN: https://developer.mozilla.org/en-US/docs/Web/API/PushManager
CanIUse: https://caniuse.com/mdn-api_pushmanager

### DIFF
--- a/feature-detects/window/pushmanager.js
+++ b/feature-detects/window/pushmanager.js
@@ -3,7 +3,7 @@
   "name": "PushManager",
   "property": "pushmanager",
   "caniuse": "mdn-api_pushmanager",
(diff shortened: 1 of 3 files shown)


rejas:
Thanks for the Fix!

Issue #2747, opened 2024-09-17
Test for flex-gap fails on Safari if zoom is 50% or 150%


jnehlmeier:
**Type of bug**

- [ ] Differences between Can I use & Modernizr
- [ ] False positive
- [x] False negative
- [ ] Modernizr error

**Describe the bug**

**How to Reproduce**

macOS 14.x, Safari 17.x, Safari 18.x

- Zoom to 50% or 150% and [run the test](https://github.com/Modernizr/Modernizr/blob/master/feature-detects/css/flexgap.js)
- For both zoom levels `flex.scrollHeight` returns 2 instead of 1 so the test fails even though Safari supports flex-gap.

**Expected behavior**

This is likely a Safari bug but the test should workaround it as the issue exists in Safari 17 and 18 (current) and possibly in earlier versions as well.

TusharNaugain:
I can resolve this by modifying the existing test to handle Safari's behavior under different zoom settings, ensuring accurate detection of flex-gap support. Given my experience with JavaScript and CSS properties, I'm well-equipped to tackle this issue effectively.

PRAteek-singHWY:
@jnehlmeier please assign this issue to me, i would love to work on it .

PR #2749, opened 2024-10-08
Update flexgap.js

TusharNaugain:
Added a check for Safari bug:

var isSupported = flexHeight === 1 || flexHeight === 2;
Reason: Safari, at zoom levels of 50% and 150%, returns scrollHeight as 2 instead of 1 for flex containers. This line ensures the test works across different browsers, including buggy Safari versions.

Fixes https://github.com/Modernizr/Modernizr/issues/2747


rejas:
This looks very wrong, are you sure this is what you want to commit?

TusharNaugain:
> This looks very wrong, are you sure this is what you want to commit?

No brother. This version of the code ensures that the flexgap test works properly in Safari, even when zoomed, by checking both possible values for scrollHeight.

rejas:
No, you first committed this code:

grafik

Can you explain why?

TusharNaugain:
@rejas could you please mention the label of hactoberfest and GSSSoc

PR #2753, opened 2024-10-24
Feature Detection Enhancements

Ansh74284:
a feature detects for the Intersection Observer API, which allows developers to asynchronously observe changes in the intersection of a target element with an ancestor element or with a top-level document's viewport. let's improve an existing feature detect for the fetch API. We can enhance the detection logic to check for the presence of the fetch function and ensure it is a function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,27 +132,18 @@ To execute the tests using mocha-headless-chrome on the console run:
 npm test
 ```
 
-You can also run tests in the browser of your choice with this command:
+You can also run tests in your browser of choice with this command:
 
 ```shell
 npm run serve-gh-pages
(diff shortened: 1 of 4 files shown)


rejas:
Thanks but we already have feature detects for these two things:
- https://github.com/Modernizr/Modernizr/blob/a747f33b2bc5db186f18bca4d1d63bf503d3159d/feature-detects/dom/intersection-observer.js
- https://github.com/Modernizr/Modernizr/blob/a747f33b2bc5db186f18bca4d1d63bf503d3159d/feature-detects/network/fetch.js

Please check before starting work on something.

PR #2776, opened 2025-07-14
fix(aspectratio): should use CSS.supports only when CSS is an object, isn't it?

HelKyle:


HelKyle:
@rejas OK, code updated

rejas:
@HelKyle I took the liberty to rework the usage of css supports in your PR. what do you think?

HelKyle:
> @HelKyle I took the liberty to rework the usage of css supports in your PR. what do you think?

Feel free to make any improvements you find necessary.

